PR #6, opened 2023-02-22
Add basic query result sourcing for core/query

alecgeatches:
## Description

(WIP) Support `core/query` block data by providing query results in JSON.

### DIFF
--- a/parser/block-additions/core-query.php
+++ b/parser/block-additions/core-query.php
@@ -1,0 +1,47 @@
+<?php
+
+namespace WPCOMVIP\ContentApi\ContentParser\BlockAdditions;
+
+use WP_Query;
+
+defined( 'ABSPATH' ) || die();
+
+class CoreQuery {
+	public static function init() {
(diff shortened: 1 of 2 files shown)


alecgeatches:
I'm going to close this branch. The purpose of this PR was to test if server-side blocks like `core/query` were supported with content API filters. We showed that this was possible with `vip_content_api__sourced_block_result` and `build_query_vars_from_query_block()` to provide results in client-side attributes.

Although this implementation may be useful, it adds additional query time to results utilizing `core/query` blocks and [hardcodes metadata like the `$page` parameter](https://github.com/Automattic/vip-content-api/blob/44f4cb5ee95a71e7201073e8bc4143fbcfcbce07/parser/block-additions/core-query.php#L35). Since we already have a `core/image` example for using the block result filter and this isn't fully fleshed out, it makes sense to remove it for now. Customers who'd like to use this functionality for core or custom blocks can use implementations specific to their code.

If this behavior is desired by default in the future, we can revisit and implement a robust wrapper for query blocks.

PR #13, opened 2023-03-14
Create issue template

ingeniumed:
### Description

This is a basic version of the issue template that customers can use to open bugs.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -23,14 +23,14 @@ A clear and concise description of what actually happened.
 **Screenshots**
 Add screenshots of the JSON response to help explain your problem. Please ensure confidential information is removed.
 
+**Block Data API URL**
+
+If publicly accessible, please share the block data API URL used to reproduce this problem.
+
+e.g. `https://my.site/wp-json/vip-block-data-api/v1/posts/139/blocks`
+
 **Version of the plugin**
 Version of the plugin
-
-**Version of WordPress**
-Version of WordPress
-
-**Version of PHP**
-PHP Version
 
 **Additional context**
 Add any other context about the problem here.
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,0 +1,36 @@
+---
+name: Bug report
+about: Create a report to help us improve
+title: ''
+labels: bug
+assignees: ingeniumed, smithjw1, alecgeatches
+
+---
+
+**Describe the bug**
+A clear and concise description of what the bug is.
+
+**To Reproduce**
+- Include what kind of a post has caused this problem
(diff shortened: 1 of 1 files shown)


alecgeatches (approved):

alecgeatches:
It would be great to get the exact link to the block data API if possible, e.g.

```
http://content-api.vipdev.lndo.site/wp-json/vip-content-api/v1/posts/261/blocks
```

This URL will likely be publicly accessible (and is by default), so this won't leak any internal information unless they choose to lock down access to the API. This will also give us enough information to figure out the site ID and post content related to the issue, which is pretty much everything we'd need to debug. How about this as a section?

```
**Block Data API URL**

If publicly accessible, can you share the block data API URL used to reproduce this problem?

e.g. `https://my.site/wp-json/vip-block-data-api/v1/posts/139/blocks`
```

ingeniumed:
> It would be great to get the exact link to the block data API if possible, e.g.
> 
> ```
> http://content-api.vipdev.lndo.site/wp-json/vip-content-api/v1/posts/261/blocks
> ```
> 
> This URL will likely be publicly accessible (and is by default), so this won't leak any internal information unless they choose to lock down access to the API. This will also give us enough information to figure out the site ID and post content related to the issue, which is pretty much everything we'd need to debug. How about this as a section?
> 
> ```
> **Block Data API URL**
> 
> If publicly accessible, can you share the block data API URL used to reproduce this problem?
> 
> e.g. `https://my.site/wp-json/vip-block-data-api/v1/posts/139/blocks`
> ```

Went back and forth on this one, because I wasn't sure if they would be comfortable revealing their details. But I do think its a great idea and would save us time and energy in debugging, so have added it in. We can ensure that we change this once we officially release the plugin, so no sensitive details are leaked.

I've also removed the WordPress and PHP versions, since we can figure that out based on their site ID.

smithjw1:
Does a screenshot of JSON make sense if we are getting the URL? Could it be different than the URL they provide?

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,36 @@
...
+- Include what kind of a post has caused this problem
+- Include any custom blocks that have caused this problem
+
+**Expected behavior**
+A clear and concise description of what you expected to happen.
+
+**Actual behavior**
+A clear and concise description of what actually happened.
+
+**Screenshots**
```
smithjw1: I'm not so sure this is necessary.

smithjw1 (approved):

.github/ISSUE_TEMPLATE/bug_report.md
ingeniumed: My main motivation for that was just in case it was intermittent or some kind of a weird bug that we weren't able to replicate. We would at least have some kind of a screenshot or snapshot of what the response was like. 

We also had reservations about customers willingly sharing that, so def do think that's an issue. I'll take it out to be safe.

PR #15, opened 2023-03-17
Update README.md

smithjw1:
Just a few changes for clarity.

## Description

A few sentences describing the overall goals of the Pull Request.

Should include any special considerations, decisions, and links to relevant GitHub issues.

## Steps to Test

Outline the steps to test and verify the PR here.

Example:

1. Check out PR.
1. Run `npm run build`.
1. Test the thing.`
1. Verify cookies are delicious.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,6 +8,16 @@
 
 A REST API to retrieve block editor posts structured as JSON data. While primarily designed for use in decoupled WordPress, the block data API can be used anywhere you want to represent block markup as structured data.
 
+## Quickstart
+
+You can get started with the Block Data API in less than five minutes.
+
+ 1. Install the plugin, we recommend using a [git subtree](#install-via-git-subtree).
+ 2. Activate the plugin
+ 3. Make a request to `/wp-json/vip-block-data-api/v1/posts/<post_id>/blocks`
+ 
+Read on for other installation options, examples, and helpful filters that you can use to customize the API to your particular use case.
+
 ## Table of contents
 
 - [Installation](#installation)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -656,7 +658,7 @@ Used to limit which post IDs are valid in the REST API. By default, all posts wi
 return apply_filters( 'vip_block_data_api__rest_validate_post_id', $is_valid, $post_id );
 ```
 
-For example, this filter can be used to allow only published `page` types to be available:
+For example, this filter can be used to allow only pages that are published to be available:
```
smithjw1, line 669: When I read this the first time, I was confused. "Right above, it said published posts are the default filter?"

It took a few readings to understand the primary change in the example is only pages are returned. 

So I flipped the order. I also just called them pages because I think most WP folks know what that means. It is less exact, but I think is better in this case.

README.md
```diff
@@ -756,6 +758,8 @@ Some frontend JavaScript frameworks require image dimensions for responsive imag
 }
 ```
 
+This filter is automatically called on image blocks.
```
smithjw1: I wasn't 100 percent sure if this was true. Since we used it as an example, it wasn't 100 percent clear.

I suggest reworking the example into something else if there is something obvious. Then you can add a paragraph that says we've implemented a default image filter if you want to see another example.

ingeniumed (approved):

README.md
alecgeatches: Good thinking here. I agree that this was a verbose example that didn't add much to the understanding of the filter. I reorganized this section to remove the Block additions/Custom block additions headers, use the short custom block code as a primary filter example, and linked to the image block code as you suggested.

alecgeatches:
@smithjw1 Thank you for the feedback! I've made a handful of changes and fixes from your suggestions. I'm going to merge this in now, but please take a look at the final changeset in this PR when you have time.

PR #17, opened 2023-03-20
Add README section on WPVIP caching

alecgeatches:
Add section explaining default caching provided on WPVIP and warnings for using authenticated requests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -749,13 +749,9 @@ For another example of how this filter can be used to extend block data, we've i
 
 ## Caching on WPVIP
 
-Requests to the block data API on WPVIP will [automatically be cached][wpvip-page-cache-cached]:
+All requests to the block data API on WPVIP will automatically be cached for 1 minute. Note that authenticated requests will bypass this cache, and so care has to be taken while using the [REST permissions filter](#vip_block_data_api__rest_permission_callback).
 
-> All [WordPress REST API][wordpress-rest-api] responses are cached for 1 minute.
-
-Although the block data API is designed to be fast when uncached, we recommend relying on the default cache timeout to reduce server load and improve performance.
-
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -744,6 +747,16 @@ Direct block HTML can be accessed through `$block['innerHTML']`. This may be use
 
 For another example of how this filter can be used to extend block data, we've implemented a default image block filter in [`parser/block-additions/core-image.php`][repo-core-image-block-addition]. This filter is automatically called on `core/image` blocks to add `width` and `height` attributes to image block attributes.
 
+## Caching on WPVIP
+
+Requests to the block data API on WPVIP will [automatically be cached][wpvip-page-cache-cached]:
+
+> All [WordPress REST API][wordpress-rest-api] responses are cached for 1 minute.
+
+Although the block data API is designed to be fast when uncached, we recommend relying on the default cache timeout to reduce server load and improve performance.
```
ingeniumed: I feel like this is trying to re-word what we already have in our docs, which are pretty detailed already. I would re-word this caching whole section to just:

All requests to the block data API on WPVIP will be automatically cached for 1 minute. Note that authenticated requests will bypass this cache, and so care has to be taken while using the [REST permissions filter](#vip_block_data_api__rest_permission_callback).

More information on this can be found [here](https://docs.wpvip.com/technical-references/caching/page-cache).

ingeniumed: Further thought - you can actually move this to where the permission filter bit is and just mention using this would break caching on WPVIP.

alecgeatches: I agree that it's not worth rephrasing WPVIP's documentation, so pushed a change to simplify things. I'm still including the caching section separately since it's still helpful information unrelated to the filter, but there's less repetition now.

PR #20, opened 2023-03-30
Fix error analytics and add local analytics

alecgeatches:
## Description

This PR addresses a few analytics-related issues:

1. Change plugin to no longer require [VIP's `vip-go-mu-plugins`](https://github.com/Automattic/vip-go-mu-plugins/).
1. Originally we added analytics that only record data when launched on WPVIP production sites. However, since the plugin is currently in a beta/"use with caution" status, this meant we were excluding all usage information in the beta. Change usage analytics to work in non-WPVIP environments.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,8 @@
 vip-block-data-api*.zip
+
+# Ignore non-production composer dependencies
+vendor/*
+!vendor/autoload.php
+!vendor/composer/
+!vendor/masterminds/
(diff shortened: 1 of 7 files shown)


alecgeatches:
Note that the test failure here is [an unrelated `wp-env`/WordPress tag issue](https://github.com/WordPress/gutenberg/issues/49453).

ingeniumed (approved):

PR #27, opened 2023-04-14
Add an allow list and block list filter for the block types in a post

ingeniumed:
## Description

This PR is meant to introduce the concept of a filter to the rest endpoint. This can act as an allow list or a block list for the block types allowed in a post. Only the allow list or block list is allowed in for filtering purposes.

## Steps to Test

- Spin up an env using `wp-env start`
- Call `https://localhost:8888/wp-json/vip-block-data-api/v1/posts/1/blocks?skip_blocks=core/paragraph` to get an empty result
- Call `https://localhost:8888/wp-json/vip-block-data-api/v1/posts/1/blocks?include_blocks=core/paragraph` to get the usual result

## TODO

- Add a test
- Update the readme with this new filter

Just want to get an initial approval on the design and then I'll bake that in.


src/parser/content-parser.php
```diff
@@ -14,6 +14,8 @@ class ContentParser {
 	protected $block_registry;
 	protected $post_id;
 	protected $warnings = [];
+	protected $blocks_to_include = [];
```
alecgeatches: I think we can skip adding these to the class because they're only used in `parse()`. That'll help keep the state small and easier to manage.

src/parser/content-parser.php
```diff
@@ -32,9 +34,11 @@ public function __construct( $block_registry = null ) {
 	 *
 	 * @return array|WP_Error
 	 */
-	public function parse( $post_content, $post_id = null ) {
+	public function parse( $post_content, $post_id = null, $blocks_to_skip = [], $blocks_to_include = [] ) {
```
alecgeatches: How about putting these into a combined parameter? I'm picturing something like this:

```php
public function parse( $post_content, $post_id = null, $options = [] ) {
    // ...
    $blocks_to_include = $options[ 'include' ] ?? [];
    $blocks_to_skip = $options[ 'exclude' ] ?? [];
```

This cleans up the method signature a bit and will make it easier to add more options without making the method signature blow up.

src/parser/content-parser.php
```diff
@@ -32,9 +34,11 @@ public function __construct( $block_registry = null ) {
...
 		$this->post_id  = $post_id;
 		$this->warnings = [];
+		$this->blocks_to_include = $blocks_to_include;
```
alecgeatches: What do you think about adding a code filter here as well?

```php
$blocks_to_include = apply_filters( 'vip_block_data_api__include_blocks', $blocks_to_include );
```

This would give another useful way to define the include/exclude list in code, e.g.

```php
add_filter( 'vip_block_data_api__include_blocks', 'setup_included_blocks' );

function setup_included_blocks( $blocks_to_include ) {
    // Only allow heading, paragraph, and my-custom-block in API results
    return [
        'core/heading',
        'core/paragraph',
        'wpvip/my-custom-block',
    ];
}
```

I think this approach would be easier if there's a big list of included or excluded blocks and a GET request is getting unwieldy. The customer can choose if the client or server decides the blocks, which is nice.

src/rest/rest-api.php
```diff
@@ -55,6 +55,13 @@ public static function permission_callback() {
 	}
 
 	public static function get_block_content( $params ) {
+		if ( isset( $params[ 'skip_blocks' ] ) && isset( $params[ 'include_blocks' ] ) ) {
+			return new WP_Error( 'vip-block-data-api-invalid-params', 'Cannot use both skip_blocks and include_blocks parameters', [ 'status' => 400 ] );
```
alecgeatches: `$content_parser->parse()` is also capable of returning `WP_Error` results (like `vip-block-data-api-no-blocks`). It might make sense to move this return into the `ContentParser` class so that this error is returned if the class is called directly.

src/rest/rest-api.php
```diff
@@ -55,6 +55,13 @@ public static function permission_callback() {
...
+		}
+
+		$blocks_to_skip = isset( $params[ 'skip_blocks' ] ) ? explode( ",", $params[ 'skip_blocks' ] ) : [];
```
alecgeatches: Make sure to also set up validate/sanitize functions for these parameters in the `register_rest_route()` call above under `'args'`.

alecgeatches (changes requested):
This is a good addition! Added a few comments above with some suggestions.

src/parser/content-parser.php
ingeniumed: Yip, agreed. Removed

src/parser/content-parser.php
ingeniumed: Yip, agreed. Done that

src/rest/rest-api.php
ingeniumed: Yip, agreed. Done

src/rest/rest-api.php
ingeniumed: Yip, agreed. Done

src/parser/content-parser.php
ingeniumed: Got that in, but the apply_filter part is broken in my commit. Working through that.

ingeniumed:
Good to review once more!

README.md
```diff
@@ -716,6 +807,26 @@ add_filter( 'vip_block_data_api__rest_permission_callback', function( $is_permit
...
+ *
+ * @param array  $is_block_included A boolean value where true means the block is included, and false to filter it out.
+ * @param string $block_name    The name of the parsed block, e.g. 'core/paragraph'.
+ * @param string $block         The result of parse_blocks() for this block.
+ *                              Contains 'blockName', 'attrs', 'innerHTML', and 'innerBlocks' keys.
+ */
+apply_filters( 'vip_block_data_api__content_filter_block', $is_block_included, $block_name, $block );
+```
+
+This is useful when it's necessary to filter out the blocks given back by the Block Data API, including in the inner blocks of a block as well.
```
alecgeatches: Can we add a couple of examples here for implementing a basic allow and blocklist via this filter?

src/parser/content-parser.php
```diff
@@ -26,13 +26,29 @@ public function __construct( $block_registry = null ) {
...
+	public function should_block_be_included($block, $block_name, $filter_options ) {
+		$is_block_included = true;
+
+		if ( ! empty( $filter_options[ 'include' ] ) ) {
+			$is_block_included = in_array( $block_name, $filter_options[ 'include' ] );
+		} else if ( ! empty( $filter_options[ 'exclude' ] ) ) {
+			$is_block_included = ! in_array( $block_name, $filter_options[ 'exclude' ] );
+		}
+
+		return apply_filters( 'vip_block_data_api__content_filter_block', $is_block_included, $block_name, $block);
```
alecgeatches: How about `vip_block_data_api__allow_block` for this filter name? I think that makes it clearer that returning `false` removes a block and `true` allows it through.

alecgeatches: Can you include the PHPDoc header from the README above this filter in code as well?

tests/rest/test-rest-api.php
```diff
@@ -264,6 +407,11 @@ public function test_rest_api_returns_error_for_classic_content() {
 		$this->expectException( \PHPUnit\Framework\Error\Error::class );
 
 		$request  = new WP_REST_Request( 'GET', sprintf( '/vip-block-data-api/v1/posts/%d/blocks', $post_id ) );
+		$request->set_query_params( array(
```
alecgeatches: Is this supposed to be in here?

tests/rest/test-rest-api.php
```diff
@@ -276,6 +424,25 @@ public function test_rest_api_returns_error_for_classic_content() {
 		wp_delete_post( $post_id );
 	}
 
+	public function test_rest_api_returns_error_for_include_and_exclude_filter() {
+		$post_id = $this->get_post_id_with_content( '<p>Classic editor content</p>' );
```
alecgeatches: We should ensure the post content is valid Gutenberg blocks, or it'll might hit the classic content error from the test above.

```suggestion
		$post_id = $this->get_post_id_with_content( '<!-- wp:paragraph --><p>content</p><!-- /wp:paragraph -->' );
```

src/parser/content-parser.php
```diff
@@ -26,13 +26,29 @@ public function __construct( $block_registry = null ) {
...
+	}
+
 	/**
 	 * @param string $post_content HTML content of a post.
 	 * @param int|null $post_id ID of the post being parsed. Required for blocks containing meta-sourced attributes and some block filters.
 	 *
 	 * @return array|WP_Error
 	 */
-	public function parse( $post_content, $post_id = null ) {
+	public function parse( $post_content, $post_id = null, $filter_options = [] ) {
```
alecgeatches, line 58: Can you add `$filter_options` to the PHPDoc here? Something like

```php
/**
 * @param string $post_content HTML content of a post.
 * @param int|null $post_id ID of the post being parsed. Required for blocks containing meta-sourced attributes and some block filters.
 * @param array $filter_options An associative array of options for filtering blocks. Can contain keys:
 *              'exclude': An array of block names to block from the response.
 *              'include': An array of block names that are allowed in the response.
 *
 * @return array|WP_Error
 */
public function parse( $post_content, $post_id = null, $filter_options = [] ) {
```

src/parser/content-parser.php
```diff
@@ -98,9 +116,13 @@ public function parse( $post_content, $post_id = null ) {
 	 *
 	 * @return array[string]array
 	 */
-	protected function source_block( $block, $registered_blocks ) {
+	protected function source_block( $block, $registered_blocks, $filter_options ) {
```
alecgeatches, line 130: Minor: Since we're changing this function to return `null` sometimes, we should include that in the PHPDoc:

```php
/*
 * ...
 * @return array[string]array|null
 */
```

alecgeatches (changes requested):
Pushed some minor README changes and rearranging for the table of contents, and left a few comments above!

tests/rest/test-rest-api.php
ingeniumed: Nope, it's a merge conflict that got resolved incorrectly since the two tests were one after the other. Good catch

tests/rest/test-rest-api.php
ingeniumed: Merge conflict side effect unfortunately as I had the Gutenberg blocks in there previously. Good catch

alecgeatches (approved):
Looks great! I'll merge this into the `v2.1.0` release.

PR #35, opened 2023-06-28
Add "Install on WordPress VIP" section to README

alecgeatches:
## Description

Add a section about WordPress VIP activation via mu-plugins to the README.

Note that the `wpvip-mu-plugins-block-data-api` link may need to be updated to match the published documentation URL. This is currently set to the non-functional link https://docs.wpvip.com/technical-references/vip-go-mu-plugins/block-data-api-plugin/.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,6 +23,7 @@ Other installation options, examples, and helpful filters for customizing the AP
 ## Table of contents
 
 - [Installation](#installation)
+	- [Install on WordPress VIP](#install-on-wordpress-vip)
 	- [Install via `git subtree`](#install-via-git-subtree)
 	- [Install via ZIP file](#install-via-zip-file)
 	- [Plugin activation](#plugin-activation)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -57,10 +58,30 @@ Other installation options, examples, and helpful filters for customizing the AP
...
+### Install on WordPress VIP
+
+The Block Data API plugin is authored and maintained by [WordPress VIP][wpvip], and made available to all WordPress sites by [VIP MU plugins][vip-go-mu-plugins]. Customers who host on WordPress VIP or use [`vip dev-env`](https://docs.wpvip.com/how-tos/local-development/use-the-vip-local-development-environment/) to develop locally have access to the Block Data API automatically. We recommend this activation method for WordPress VIP customers.
+
+Enable the plugin by adding the method shown below to your application's [`client-mu-plugins/plugin-loader.php`][vip-go-skeleton-plugin-loader-example]:
+
+```php
+// client-mu-plugins/plugin-loader.php
+
+\Automattic\VIP\Integrations\activate( 'block-data-api' );
```
smithjw1, line 73: I made this comment in the docs as well. Do we need to be more defensive about this? Checking to make sure the integration exists before we call it?

smithjw1 (approved):

README.md
alecgeatches: Interesting idea. Currently, the only public API method we expose is [`activate()`](https://github.com/Automattic/vip-go-mu-plugins/blob/develop/vip-integrations.php#L29-L32) used above. If the integration doesn't exist, [we throw an error here](https://github.com/Automattic/vip-go-mu-plugins/blob/ae85fa8/integrations/integrations.php#L86) essentially as a typo check. I could see a couple of ways we could address your comment:

1. Don't throw an error at all, or trigger a lesser PHP warning instead.
2. Add a `if_integration_available()` method to allow for safely checking that a slug exists.

However, this would really only affect client code if we remove or change the `block-data-api` slug in the future. Right now this will work for any site with access to mu-plugins/the `activate()` function.

I'm not sure that we need an additional check. If we make slug name changes in the future, I think we'd likely use a layer to convert slugs to a newer correct version. If we remove an existing integration, we can make the PHP warning changes to avoid breaking client sites. That being said, if we do remove an integration, we'd still need to alert customers so that site dependencies don't suddenly disappear. If `block-data-api` stops working, I think we'll need to address this in other ways beyond a flag check.

smithjw1: I also thinking, what if I take this code and move it to somewhere other than VIP, or if there's a weird deployment and the `\Automattic\VIP\Integrations\` class doesn't exist?

Maybe too defensive. Happy to hear that.

alecgeatches: > I also thinking, what if I take this code and move it to somewhere other than VIP

That's a fair point. In that case, it'll fail right away, which it probably should as a feature that relies on VIP code. The `client-mu-plugins/` folder the configuration lives in is also a VIP thing, so it's even more likely the whole folder would just be ignored outside of VIP.

> or if there's a weird deployment and the `\Automattic\VIP\Integrations\` class doesn't exist?

That could happen, but then it's difficult to have an API check since we don't have a function to call. A super defensive activation using an imaginary `if_integration_exists()` might look like this:

```php
if ( 
    function_exists('\Automattic\VIP\Integrations\if_integration_exists' ) &&
    function_exists('\Automattic\VIP\Integrations\activate' ) &&
    \Automattic\VIP\Integrations\if_integration_exists( 'block-data-api' ) ) {
    \Automattic\VIP\Integrations\activate( 'block-data-api' );
}
```

or, with the current setup:

```php
if ( function_exists('\Automattic\VIP\Integrations\activate' ) ) {
    try {
        \Automattic\VIP\Integrations\activate( 'block-data-api' );
    } catch ($e) {
        // ...
    }
}
```

I'd still think this isn't necessary though. If the `activate()` function doesn't exist in some deployment, and a site relies on the integration, something should probably go wrong.

ingeniumed (approved):

PR #37, opened 2023-07-19
Add API versioning plans to README

alecgeatches:
## Description

As we're now shipping with [VIP Integrations](https://docs.wpvip.com/technical-references/vip-go-mu-plugins/block-data-api-plugin/) for managing plugin updates, include a section in the README about our API versioning plans.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,6 +30,7 @@ Other installation options, examples, and helpful filters for customizing the AP
 	- [Install via ZIP file](#install-via-zip-file)
 	- [Plugin activation](#plugin-activation)
 - [Usage](#usage)
+ 	- [Versioning](#versioning)
 - [Block Data API examples](#block-data-api-examples)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -142,6 +145,16 @@ Review these [**Filters**](#filters) to learn more about limiting access to the
...
 
+### Versioning
+
+The current REST endpoint uses a `v1` prefix:
+
+```
+/wp-json/vip-block-data-api/v1/...
+```
+
+We plan to utilize API versioning to avoid unexpected changes to the plugin. In the event that we make breaking changes to API output, we will add a new endpoint (e.g. `/wp-json/vip-block-data-api/v2/`) with access to new data. We intend to keep backwards compatibility with existing version endpoints.
```
smithjw1: "existing version endpoints." reads oddly to me, though I don't know why.

"Previous versions will remain accessible for backward compatibility."

alecgeatches: Much better, thank you!

Issue #39, opened 2023-07-27
Privacy Analytics on by default for everybody and can't be turned off


tomjn:
**Describe the bug**

The plugin phones home on every use, doesn't share a privacy policy, uses a wp.com tracking pixel to perform this with no mention of it in the readme, and provides no constants/filters/settings to control this.

**To Reproduce**

Use the plugin

**Expected behavior**

 - A privacy policy or link to a privacy policy ( the wp.com tracking pixel privacy policy would suffice )
 - A way to turn this off

**Actual behavior**
Other than a brief mention in the readme there's no details on how it does it, where the data goes, or how to disable it

**Block Data API URL**

`*`

**Version of the plugin**

1.0.1

**Additional context**

I understand if you want this always on while used on the VIP platform where you're protected by a literal contract, that makes sense.

I do know though that although you're sending the data from PHP, protecting the visitor to the site, you're still exposing information about the host, which can be problematic. Staging environments for unlaunched sites, personal machines, etc.

Nevermind GDPR compliance, it's an unnecessary risk. _Making it opt in eliminates all the risk_, I'd settle for a simple filter or constant and a link to the wp.com privacy policy in the readme

alecgeatches:
@tomjn I agree with what you've stated above. We do have a [README Analytics section](https://github.com/Automattic/vip-block-data-api#analytics) explaining analytics use, but as you mentioned there is no opt-in/out option or privacy policy. The analytics were originally designed for beta usage (as the only identifiable information we collect is the VIP site ID, if present) and we haven't addressed this since public release. We should fix the analytics implementation.

ingeniumed:
We have turned off usage tracking for non-vip sites, as well as linked our privacy policy for vip sites in the README. We are also planning to add a filter for VIP sites to opt out of analytics as well.

As a result, this issue should be good to close?

tomjn:
It will do, if you want perfection though you should mention it uses wp.com pixels as although you/VIP only send 2 data points, wp.com might store additional data based on the IP as a result of different product decisions. That could result in data being stored that is not mentioned here.

I don't expect VIP to be able to keep up with that and babysit .com divisions privacy changes but if you link to their privacy policy you don't have to.

PR #40, opened 2023-07-27
Ability to turn off analytics for non-VIP sites

tomjn:
## Description

An example of solving #39. I expect the readme changes and filter name will need adjusting. I'd also recommend a non-american engineer look at this.

I'd also suggest differentiating further between VIP sites running locally, and VIP sites running in production. Do you really want your analytics gummed up with CI test runs and local dev environments?

### DIFF
--- a/src/analytics/analytics.php
+++ b/src/analytics/analytics.php
@@ -57,6 +57,12 @@ class Analytics {
 	}
 
 	private static function send_pixel( $stats ) {
+		$can_send = apply_filters( 'vip-block-data-api/analytics', false );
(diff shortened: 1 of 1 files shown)


ingeniumed:
@tomjn Thanks for putting up this draft PR, we appreciate it. I just put up #42 to address this. We are having a follow up discussion for opt in/out for VIP sites, but the main goal for now was to ensure that non-vip sites can have their analytics turned off.

ingeniumed:
Looks like it auto-closed this PR once I merged in #42, to address #39

PR #42, opened 2023-07-31
Remove analytics tracking for non-vip sites

### LINKED ISSUE #39: Privacy Analytics on by default for everybody and can't be turned off

ingeniumed:
## Description

This will resolve #40 by turning off analytics tracking for non-vip sites. This way if you aren't on VIP there's no opt out to be done.

Follow up discussion is in the works to see if we can allow opt out/in for VIP sites as well. Getting this in first is important to ensure non vip sites don't have analytics being sent.

## Steps to Test

1. Check out PR.
2. Spin up a wp-env and ensure nothing is sent
3. Turn off the VIP sites check and set the keys to have -test at the end
4. Spin up a vip dev-env with this and ensure stats are sent

### DIFF
--- a/README.md
+++ b/README.md
@@ -932,7 +932,7 @@ For another example of how this filter can be used to extend block data, we have
 
 ## Analytics
 
-**Please note that, this is for VIP sites only. Analytics is disabled if this plugin is not being run on VIP sites.**
+**Please note that, this is for VIP sites only. Analytics are disabled if this plugin is not being run on VIP sites.**
 
 The plugin records two data points for analytics, on VIP sites:
 
--- a/README.md
+++ b/README.md
@@ -25,40 +25,40 @@ Other installation options, examples, and helpful filters for customizing the AP
 ## Table of contents
 
 - [Installation](#installation)
-	- [Install on WordPress VIP](#install-on-wordpress-vip)
-	- [Install via `git subtree`](#install-via-git-subtree)
-	- [Install via ZIP file](#install-via-zip-file)
-	- [Plugin activation](#plugin-activation)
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -932,15 +932,15 @@ For another example of how this filter can be used to extend block data, we have
 
 ## Analytics
 
-The plugin records two data points for analytics:
+**Please note that, this is for VIP sites only. Analytics is disabled if this plugin is not being run on VIP sites.**
```
alecgeatches:
```suggestion
**Please note that this applies to VIP sites only. Analytics are disabled if this plugin is not being run on VIP sites.**
```

alecgeatches (approved):
Looks good!

Issue #41, opened 2023-07-27
PHPDocs and Typehints


tomjn:
**Describe the bug**

There's almost no type hinting in the plugin, and half the methods are missing PHPDocs.

**To Reproduce**

look at the code, I would have expected the VIP PHPCS rules to have flagged the missing PHPDocs, or WPCS. PHStan should be kicking up a fuss about it too

**Version of the plugin**
 1.0.1

ingeniumed:
@tomjn Thanks for flagging this, I agree with you on getting this added in. We will work on getting this resolved 👍🏾

ingeniumed:
PHPDocs are in, type hinting is what's left to be added in.

PR #43, opened 2023-08-01
Add PHPDocs and add it to the phpcs workflow

ingeniumed:
## Description

This will add PHPDocs everywhere, and will enable PHPDocs validation as part of the PHPCS checks. This will partially address #41.

Another PR will be incoming in which type hinting will be added. 

## Steps to Test

There should be no tests to run or changes to check because it's only docs that have been updated.

### DIFF
--- a/src/analytics/analytics.php
+++ b/src/analytics/analytics.php
@@ -1,4 +1,10 @@
 <?php
+/**
+ * Analytics
+ * 
+ * @file
+ * @package vip-block-data-api
+ */
 
 namespace WPCOMVIP\BlockDataApi;
 
@@ -7,9 +13,20 @@ defined( 'ABSPATH' ) || die();
 define( 'WPCOMVIP__BLOCK_DATA_API__STAT_NAME__USAGE', 'vip-block-data-api-usage' );
 define( 'WPCOMVIP__BLOCK_DATA_API__STAT_NAME__ERROR', 'vip-block-data-api-error' );
 
+/**
+ * Analytics Class that will be used to send data to the WP Pixel.
+ */
 class Analytics {
+	/**
+	 * Array of analytics to send to the WP Pixel.
+	 * 
+	 * @var array
+	 */
 	private static $analytics_to_send = [];
 
+	/**
+	 * Initialize the Analytics class.
+	 */
 	public static function init() {
 		add_action( 'shutdown', [ __CLASS__, 'send_analytics' ] );
 	}
--- a/src/analytics/analytics.php
+++ b/src/analytics/analytics.php
@@ -14,7 +14,10 @@ class Analytics {
 		add_action( 'shutdown', [ __CLASS__, 'send_analytics' ] );
 	}
 
-	public static function record_usage() {
+	/**
+	 * Record the usage of the plugin, for VIP sites only. For non-VIP sites, this is a no-op.
+	 */
+	public static function record_usage(): void {
 		// Record usage on WPVIP sites only
 		if ( ! self::is_wpvip_site() ) {
 			return;
@@ -24,11 +27,13 @@ class Analytics {
 	}
 
 	/**
+	 * Record an error if it's allowed, for VIP sites only. For non-VIP sites, this is a no-op.
+	 * 
 	 * @param WP_Error $error
 	 *
 	 * @return void
 	 */
-	public static function record_error( $error ) {
+	public static function record_error( $error ): void {
 		$error_data    = $error->get_error_data();
 		$error_details = isset( $error_data['details'] ) ? sprintf( ' - %s', ( $error_data['details'] ) ) : '';
 
@@ -45,7 +50,10 @@ class Analytics {
 		}
 	}
 
-	public static function send_analytics() {
+	/**
+	 * Send the analytics, if present. If an error is present, then usage analytics are not sent. 
+	 */
+	public static function send_analytics(): void {
 		if ( empty( self::$analytics_to_send ) ) {
 			return;
 		}
@@ -61,7 +69,12 @@ class Analytics {
 		self::send_pixel( self::$analytics_to_send );
 	}
 
-	private static function send_pixel( $stats ) {
+	/**
+	 * Send the stats to the WP Pixel endpoint.
+	 * 
+	 * @param array $stats
+	 */
+	private static function send_pixel( $stats ): void {
 		$query_args = [
 			'v' => 'wpcom-no-pv',
 		];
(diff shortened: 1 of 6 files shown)


src/analytics/analytics.php
```diff
@@ -58,12 +82,17 @@ public static function send_analytics() {
 			unset( self::$analytics_to_send[ WPCOMVIP__BLOCK_DATA_API__STAT_NAME__USAGE ] );
 		}
 
-		// Use the built in mu-plugins methods to send the data to VIP Stats
+		// Use the built in mu-plugins methods to send the data to VIP Stats.
```
alecgeatches:
```suggestion
		// Use built-in VIP mu-plugins method to send analytics to VIP Stats, if present.
```

src/parser/block-additions/core-image.php
```diff
@@ -1,21 +1,38 @@
...
+	 * 
+	 * @param array    $sourced_block the sourced block result.
+	 * @param string   $block_name the name of the block.
+	 * @param int|null $post_id the id of the post.
+	 * @param array    $block the block itself.
+	 * 
+	 * @access private
 	 *
-	 * @return array[string]array
+	 * @return array the updated sourced block, with the new metadata information
```
alecgeatches: Nit - I think these look better as individual capitalized "sentences", dropping the "the". For example:

```suggestion
	 * @return array Updated sourced block with new metadata information
```

Same goes for parameters, for example:

```
@param array $sourced_block Sourced block result.
```

src/rest/rest-api.php
```diff
@@ -8,15 +13,30 @@
...
+	 * Initialize the Rest API class.
+	 * 
+	 * @access private
+	 */
 	public static function init() {
 		add_action( 'rest_api_init', [ __CLASS__, 'register_rest_routes' ] );
 	}
 
+	/**
+	 * Validate the block names that have been provided, to ensure that they are valid
```
alecgeatches:
```suggestion
	 * Validate block names are non-empty and match a `<prefix>/<block-name>` naming convention.
```

src/parser/block-additions/core-image.php
ingeniumed: Yip, I get what you mean. Making that change rn

src/rest/rest-api.php
```diff
@@ -8,15 +13,30 @@
...
+	 * 
+	 * @param string $param the block names to validate.
+	 * 
+	 * @return bool true, if they are valid or false otherwise
+	 */
 	public static function validate_block_names( $param ) {
 		$block_names = explode( ',', trim( $param ) );
 
-		// Validate that all block names are valid
+		// Validate that all block names are strings and are in the format of namespace/block-name.
```
alecgeatches: Whoops, just saw this clarification after [the comment above](https://github.com/Automattic/vip-block-data-api/pull/43/files#r1281232483). If we include that information in the function comment, we can probably remove it here.

src/rest/rest-api.php
```diff
@@ -26,6 +46,9 @@ public static function validate_block_names( $param ) {
 		return true;
 	}
 
+	/**
+	 * Register the rest routes
```
alecgeatches, line 49: Doesn't seem necessary, but I do like an explicit `@return void`.

alecgeatches: Also probably `@access private` here as it's a WordPress callback not for direct use

src/analytics/analytics.php
```diff
@@ -58,12 +88,17 @@ public static function send_analytics() {
...
+		// Use built-in VIP mu-plugins method to send analytics to VIP Stats, if present.
 		if ( function_exists( '\Automattic\VIP\Stats\send_pixel' ) ) {
 			\Automattic\VIP\Stats\send_pixel( self::$analytics_to_send );
 		}
 	}
 
+	/**
+	 * Check if the site is a WPVIP site.
+	 * 
+	 * @return bool true if it is a WPVIP site, false otherwise
```
alecgeatches, line 100: Nit:
```suggestion
	 * @return bool True if it is a WPVIP site, false otherwise
```

src/parser/content-parser.php
```diff
@@ -10,13 +15,39 @@
 use WP_Block_Type_Registry;
 use Symfony\Component\DomCrawler\Crawler;
 
+/**
+ * The content parser that would be used to transform a post into an array of blocks, along with their attributes.
```
alecgeatches, line 19: I'd remove the tense here:

```suggestion
 * Content parser used to transform a post into an array of blocks and associated attributes.
```

src/parser/content-parser.php
```diff
@@ -10,13 +15,39 @@
...
+	 * 
+	 * @access private
+	 */
 	protected $warnings = [];
 
 	/**
-	 * @param WP_Block_Type_Registry|null $block_registry
+	 * Initialize the ContentParser class.
+	 *
+	 * @param WP_Block_Type_Registry|null $block_registry the block registry instance.
```
alecgeatches, line 50: Nit:

```suggestion
	 * @param WP_Block_Type_Registry|null $block_registry Block registry instance.
```

src/parser/content-parser.php
```diff
@@ -26,6 +57,20 @@ public function __construct( $block_registry = null ) {
...
+	 * - include parameter, if it is set or 
+	 * - exclude parameter, if it is set.
+	 * 
+	 * and finally, based on a filter vip_block_data_api__allow_block
+	 * 
+	 * @param array  $block Current block.
+	 * @param string $block_name Name of the block.
+	 * @param array  $filter_options Options to be used for filtering, if any.
+	 * 
+	 * @return bool true, if the block should be included or false otherwise
```
alecgeatches, line 72: Same capitalization nit.

src/parser/content-parser.php
```diff
@@ -240,7 +294,7 @@ protected function source_attribute( $crawler, $block_attribute_definition ) {
 		} elseif ( 'tag' === $attribute_source ) {
 			$attribute_value = $this->source_block_tag( $crawler, $block_attribute_definition );
 		} elseif ( 'raw' === $attribute_source ) {
-			$attribute_value = $this->source_block_raw( $crawler, $block_attribute_definition );
+			$attribute_value = $this->source_block_raw( $crawler );
```
alecgeatches, line 303: Good catch! Unused parameter.

src/parser/content-parser.php
```diff
@@ -448,16 +514,18 @@ protected function source_block_meta( $block_attribute_definition ) {
...
 	/**
-	 * @param Symfony\Component\DomCrawler\Crawler $crawler
-	 * @param array $block_attribute_definition
+	 * Helper function to process the `children` source attribute.
+	 * 
+	 * @param Symfony\Component\DomCrawler\Crawler $crawler Crawler instance.
+	 * @param array                                $block_attribute_definition Definition of the block attribute.
 	 *
-	 * @return string|null
+	 * @return array|string|null
```
alecgeatches, line 542:
```suggestion
	 * @return array
```

I think is correct here.

alecgeatches (approved):
Looks great to me! Thank you for making all these improvements.

PR #48, opened 2023-09-04
Introduce an analytics sampling rate for VIP sites

### LINKED ISSUE #45: Make analytics calls faster
Currently, the longest part of the Block Data API response is the analytics call to MC Stats. We are tracking every single request.

The goal of the analytics is to understand two things:

1. Who is using the Block Data API?
2. How is their usage changing over time?

We don't need to slow down every request to be able to answer these questions. We can use some naive sampling that will give us the data we need. Naive sampling will be fast because it doesn't have to be perfectly accurate, just directionally helpful. 

I would suggest sending the pixel at a regular time interval. For example, we could only send the pixel when the `minute` is equal to 10.  

That sort of naive sampling is useful because the use case for the read methods of the Block Data API is fairly consistent use. People will be making requests throughout the day, and we will catch those.

We would miss "burst" usage because of some job, but we could mitigate that by checking more frequently, like when the `second` equals 10. That may miss very spotty consistent usage.

Combining the two may be best. With the data we have, we should be able to determine intervals that make our code execute quickly, but still provide the directional data we need to access the use of the feature over time.

ingeniumed:
## Description

This PR will introduce a sampling algorithm for the analytics that sent, when this plugin is used on a VIP site. The sampling algorithm is as follows:

- Sample every 10s, or
- Sample every 10m.

This will ensure that we catch burst usage, or regular usage throughout the day without losing the core purpose behind the analytics.

Fixes #45 

## Steps to Test

1. Spin up a local dev using `vip dev-env`
2. Adjust the `is_wpvip_site()` to give back true
3. Pop a breakpoint inside `is_it_sampling_time`.
4. Call `http://<slug>.vipdev.lndo.site/wp-json/vip-block-data-api/v1/posts/1/blocks`, while keeping an eye on the time
5. Every 10m or 10s it should trigger the breakpoint, otherwise it shouldn't be triggered.


tests/rest/test-rest-api.php
```diff
@@ -107,7 +107,7 @@ public function test_rest_api_returns_blocks_for_post() {
 					'mediaId'           => 6,
 					'mediaLink'         => 'https://gutenberg-block-data-api-test.go-vip.net/?attachment_id=6',
 					'mediaType'         => 'image',
-					'align'             => 'wide',
+					'align'             => 'none',
```
ingeniumed, line 110: From my local testing, it looks like the default value for this has changed so this keeps it in line with that. Wonder if it's a 6.3 change?

alecgeatches: Good catch! Looks like this was [changed 6 months ago in Gutenberg](https://github.com/WordPress/gutenberg/commit/ced52228a39ba68fea86c1e3e8d36aa7947685db#diff-4d936a4cb04a450622ce03fd585affd8c3bccb3982d49faae3201ae71dbbf940L13), which was later [merged into WordPress 6.3](https://github.com/WordPress/WordPress/commit/00401e16b30c2929e8adb5e07521b3ace4dbf826) like you said.

src/analytics/analytics.php
```diff
@@ -105,6 +102,34 @@ private static function is_wpvip_site() {
...
+	 * @return boolean true, if it should be sampled or false otherwise.
+	 */
+	private static function is_it_sampling_time() {
+		$current_timestamp = current_datetime();
+
+		// Get the minutes from the date.
+		$minutes = $current_timestamp->format( 'i' );
+
+		// Get the seconds from the date.
+		$seconds = $current_timestamp->format( 's' );
```
alecgeatches: [`format()` produces strings](https://www.php.net/manual/en/datetime.format.php), so we should probably `intval()` these or expressions like `0 !== $seconds` (i.e. `0 !== "0"`) won't work as expected.

src/analytics/analytics.php
```diff
@@ -105,6 +102,34 @@ private static function is_wpvip_site() {
 			&& defined( 'FILES_CLIENT_SITE_ID' )
 			&& function_exists( '\Automattic\VIP\Stats\send_pixel' );
 	}
+
+	/**
+	 * Check if the current request should be sampled for analytics.
+	 * 
+	 * Current sampling algorithm is that every 10s or 10m, we send analytics.
+	 * 
+	 * Max calls possible based on 1 call every 10s: 8640.
```
alecgeatches: @ingeniumed Please correct me if I'm wrong, but does this math work? We could have several requests come through in the same second which would result in multiple analytics calls in the same second. I think that's fine for sampling purposes, but it doesn't necessarily give us an upper limit on analytics requests.

src/analytics/analytics.php
alecgeatches: Actually, thinking about this a bit more we may not need checks if `seconds`/`minutes` is `0`. If I'm doing my math right, the code (with proper types) would be skipping 1/6th of the sampling times:

```
00:00:00 -> Skip, seconds is 0
00:00:10 -> Send analytics
00:00:20 -> Send analytics
00:00:30 -> Send analytics
00:00:40 -> Send analytics
00:00:50 -> Send analytics
00:01:00 -> Skip, seconds is 0
```

src/analytics/analytics.php
```diff
@@ -105,6 +102,34 @@ private static function is_wpvip_site() {
...
+		$current_timestamp = current_datetime();
+
+		// Get the minutes from the date.
+		$minutes = $current_timestamp->format( 'i' );
+
+		// Get the seconds from the date.
+		$seconds = $current_timestamp->format( 's' );
+
+		// Only send analytics every 10 minutes or 10 seconds.
+		if ( ( 0 !== $seconds && 0 === $seconds % WPCOMVIP__BLOCK_DATA_API__STAT_SAMPLING_RATE_SEC ) || ( 0 !== $minutes && 0 === $minutes % WPCOMVIP__BLOCK_DATA_API__STAT_SAMPLING_RATE_MIN ) ) {
```
alecgeatches: @ingeniumed Can you explain a bit why we're doing one sample every 10 seconds and another every 10 minutes? It seems like the sampling math is easier to estimate real usage if we stick with just one of these. Even just doing seconds would reduce analytics by 90% but still give us relatively good coverage.

alecgeatches (changes requested):
Sampling is a great way to reduce our analytics overhead with no extra db/caching cost, I love it. I have a few questions above about the implementation.

src/analytics/analytics.php
ingeniumed: Yip, updated that! Good catch 👍🏾

alecgeatches (approved):

src/analytics/analytics.php
ingeniumed: You are right, I added that in mainly for reference. But I don't want to make us think that we can use this to determine actual usage easily. I've taken it out

src/analytics/analytics.php
ingeniumed: Discussed, and I have dropped the 10m leaving just 10s in place.

README.md
```diff
@@ -942,6 +942,8 @@ The plugin records two data points for analytics, on VIP sites:
 
 Both of these data points are a counter that is incremented, and do not contain any other telemetry or sensitive data. You can see what's being [collected in code here][repo-analytics], and WPVIP's privacy policy [here](https://wpvip.com/privacy/).
 
+In addition, the analytics are sent every 10 seconds only.
```
smithjw1, line 945: We could say "analytics are sampled" that way we don't have to update the README if we change the sampling rate.

src/analytics/analytics.php
```diff
@@ -105,6 +102,27 @@ private static function is_wpvip_site() {
...
+	 * @return boolean true, if it should be sampled or false otherwise.
+	 */
+	private static function is_it_sampling_time() {
+		$current_timestamp = current_datetime();
+
+		// Get the seconds from the date.
+		$seconds = intval( $current_timestamp->format( 's' ) );
+
+		// Only send analytics every 10 seconds.
+		if ( 0 === ( $seconds % WPCOMVIP__BLOCK_DATA_API__STAT_SAMPLING_RATE_SEC ) ) {
```
smithjw1, line 120: Someday we may want a hook so that customers can sent their own sampling rate no lower than X, but for a different day.

smithjw1 (approved):
Does what it says on the tin!

README.md
ingeniumed: While I do see what you mean, I would lean towards being open about the sampling rate as well as the method with which we do that in the README. That way, we are being transparent about it and when we do change it we also have to update the README to reflect that.

src/analytics/analytics.php
ingeniumed: I do see the value of that, especially if they are at a scale where even the tiniest change could have a large impact. We would need to figure out a good floor that we can use, that would still make the data useful to us. I'll add this to the backlog.

PR #51, opened 2023-11-14
Add GraphQL API

### LINKED ISSUE #49: Add in WPGraphQL support
Explore, and add in support for WPGraphQL within the plugin. This should result in a GraphQL endpoint being available, if WPGraphQL is detected. 

This should also include a filter that allows modifying the output. It'll be used by [the decoupled bundle](https://github.com/Automattic/vip-decoupled-bundle#wpgraphql-content-blocks) to maintain backwards compatibility.

ingeniumed:
## Description

This PR is meant add a GraphQL API to the Block Data API. It assumes that you have WP-GraphQL installed, and uses a filter to tweak the output of the `ContentParser` to generate the GraphQL response.

In addition, it uses the `wp_unique_id` function to generate ids for the blocks so that the blocks can be correlated together. The `inner_blocks` are therefore able to be flattened entirely, thereby bypassing the nested depth problem in graphQL.

This is in the draft status to get initial feedback, as well as to add some tests and guard rails around the GraphQL API.

graphql-rest-demo

Resolves #49 

## Steps to Test

- Download, install and activate WP-GraphQL
- Activate Block Data API with this branch
- Modify the initial post to add some nested blocks
- Run the following GraphQL query
```graphQL
query NewQuery {
  post(id: "1", idType: DATABASE_ID) {
    blocksData {
      blocks {
        id
        name
        attributes {
          name
          value
        }
        innerBlocks {
          name
          parentId
          id
          attributes {
            name
            value
          }
        }
      }
    }
  }
}
```
- Ensure the output matches the post as well as the rest API


src/graphql/graphql-api.php
```diff
@@ -0,0 +1,231 @@
...
+	/**
+	 * Extract the blocks data for a post, and return back in the format expected by the graphQL API.
+	 *
+	 * @param  WPGraphQL\Model\Post $post_model Post model for post.
+	 * @return array
+	 */
+	public static function get_blocks_data( $post_model ) {
+		$post_id        = $post_model->ID;
+		$post           = get_post( $post_id );
+		$filter_options = [ 'graphQL' => true ];
```
ingeniumed: Using this as a way to signify that it's a graphQL request and so the filter would only target those calls.

src/graphql/graphql-api.php
```diff
@@ -0,0 +1,231 @@
...
+ * GraphQL API to offer an alternative to the REST API.
+ */
+class GraphQLApi {
+	/**
+	 * Initiatilize the graphQL API, if its allowed
+	 *
+	 * @access private
+	 */
+	public static function init() {
+		$is_graphql_to_be_enabled = apply_filters( 'vip_block_data_api__is_graphql_enabled', true );
```
ingeniumed: True by default, but can be set to false if need be.

src/graphql/graphql-api.php
```diff
@@ -0,0 +1,231 @@
...
+		$filter_options = [ 'graphQL' => true ];
+
+		$content_parser = new ContentParser();
+
+		$parser_results = $content_parser->parse( $post->post_content, $post_id, $filter_options );
+
+		// ToDo: Verify if this is better, or is returning it another way in graphQL is better.
+		if ( is_wp_error( $parser_results ) ) {
+			// Return API-safe error with extra data (e.g. stack trace) removed.
+			return new \Exception( $parser_results->get_error_message() );
```
ingeniumed, line 47: Throwing WP_Error caused problems in WP-GraphQL, which is why I've done this.

src/parser/content-parser.php
```diff
@@ -233,10 +233,16 @@ protected function source_block( $block, $registered_blocks, $filter_options ) {
 		$sourced_block = [
 			'name'       => $block_name,
 			'attributes' => $block_attributes,
+			'id'         => wp_unique_id(),
```
ingeniumed: This was a neat discovery as the IDs are unique, albeit different each time a request is made. The advantage is that we can use this to setup parent-child relationships at least.

src/parser/content-parser.php
```diff
@@ -233,10 +233,16 @@ protected function source_block( $block, $registered_blocks, $filter_options ) {
...
 		];
 
+		if ( isset( $filter_options['parentId'] ) ) {
+			$sourced_block['parentId'] = $filter_options['parentId'];
+		}
+
 		if ( isset( $block['innerBlocks'] ) ) {
-			$inner_blocks = array_map( function ( $block ) use ( $registered_blocks, $filter_options ) {
+			$filter_options['parentId'] = $sourced_block['id'];
```
ingeniumed: The ID of a block is set as the parent ID on its children by passing it down in the filter itself.

src/parser/content-parser.php
```diff
@@ -256,13 +262,26 @@ protected function source_block( $block, $registered_blocks, $filter_options ) {
...
+		/**
+		 * Filters a block when parsing is complete.
+		 *
+		 * @param array $sourced_block An associative array of parsed block data with keys 'name' and 'attribute'.
+		 * @param string $block_name Name of the parsed block, e.g. 'core/paragraph'.
+		 * @param int $post_id Post ID associated with the parsed block.
+		 * @param array $block Result of parse_blocks() for this block. Contains 'blockName', 'attrs', 'innerHTML', and 'innerBlocks' keys.
+		 * @param array $filter_options Options to filter using, if any.
+		 */
+		$sourced_block = apply_filters( 'vip_block_data_api__sourced_block_result_transform', $sourced_block, $block_name, $this->post_id, $block, $filter_options );
```
ingeniumed: Made a new filter because the previous one had an extra param added to it, so it's a breaking change. This avoids that.

src/graphql/graphql-api.php
```diff
@@ -0,0 +1,231 @@
...
+					],
+				],
+			],
+		);
+
+		// Register the type corresponding to the individual block, with the above attribute.
+		register_graphql_type(
+			'BlockData',
+			[
+				'description' => 'Block data',
```
alecgeatches, line 178: @ingeniumed Is there a way we can combine `BlockData` and `InnerBlockData` so that they're one datatype? It seems more intuitive to me that there's a single type, where a root `BlockData` has an empty parent.

alecgeatches: One more follow-up: am I correct that we're returning a flattened array of blocks that can be reconstructed by following `id`/`parentId` references? If that's the case, can we leave `InnerBlockData`/the `innerBlocks` attribute out of the GraphQL definitions, and rely on post-processing to make a block tree instead? It seems like there's a hybrid going on right now that feels unusual.

src/parser/content-parser.php
alecgeatches: Because this is additive, I don't think this is a breaking update. Implementers of the filter pass an explicit number of accepted parameters, like `5` in your usage here:

https://github.com/Automattic/vip-block-data-api/blob/778509d781453d693e368a19e6ac71f52c99e186/src/graphql/graphql-api.php#L28

Prior users of the filter would register for up to `4` arguments, which would continue to work here as the newest parameter is added to the end. [The docs page for `add_filter()`](https://developer.wordpress.org/reference/functions/add_filter/) sort of touches on that, but you can test that using `4` (or any lower positive number) instead of `5` continues to work without a break.

src/parser/content-parser.php
alecgeatches: Do these IDs show up in the REST response as well? If so, can we make it so that we only assign IDs for the GraphQL response? It's not a huge addition of data, but I don't want to add unused metadata to our REST response if it's already in a hierarchical format.

alecgeatches: Maybe it'd be cleaner to do ID assigning as a post-processing step in `flatten_inner_blocks()`. I think that makes sense, as the parent/child ID relationship is a product of the flattening and GraphQL limitations, not necessarily inherent to the data.

ingeniumed: >It's not a huge addition of data, but I don't want to add unused metadata to our REST response if it's already in a hierarchical format.
> Maybe it'd be cleaner to do ID assigning as a post-processing step in flatten_inner_blocks().

I should have expanded the reasoning a bit more in my comment.

My main reasoning for keeping this in here, rather than in just the graphQL API was to account for unique IDs playing a role down the line. At the moment, these IDs are generated each time a request is sent. But, what I would like to do is to make the IDs get generated just once and then be able to re-use them. This way, you don't just need to query all the blocks but could actually query a specific block. This could be useful down for a variety of use cases. In addition to that, if the Gutenberg/core folks decide to make block IDs unique we would be able to leverage them here as well.

So this allows us to set that system up. Hope that makes a lot more sense as to why it's done this way?

src/graphql/graphql-api.php
ingeniumed: That's correct. 

The way the schema is:

- The outer list is the list of root blocks
- Within each root block there is a flattened innerBlocks list, using a `id/parentId` relationship to sort out how the innerBlocks would be reconstructed

The reason why the schema has two different data types for inner blocks, and outer blocks is because we would want the inner blocks to be picked only once. Otherwise, at n depths the inner blocks can be picked as a possible field which isn't right. The current way ensures that it can only be picked one - at depth 1. Check out [this](https://github.com/Automattic/vip-block-data-api/pull/51/commits/4c00b0d2b4f3d6dcb7e70b7bc0ef8267a4c85b38) commit for how it was done before, as that's the way that you are suggesting if I'm understanding it right.

So I would leave this implementation the way it is.

Lemme know if what I said makes sense 👍🏾

src/parser/content-parser.php
ingeniumed: Fair point, I've made that change. It's down to just the existing filter being modified 👍🏾

src/parser/content-parser.php
ingeniumed: I decided to give this a shot, and I almost have it working. There's repeated outer inner blocks that keep showing up, but otherwise it's working. Once I fix that bug I'll push it up

ingeniumed: Got it working! It's now only going to be present in the GraphQL, and the side effect of this is that in the filter_options we pass the current id and parent_id of the block being worked on. That should set us up for the future, for the case I mentioned while cleaning up the rest response.

alecgeatches:
@ingeniumed I noticed some errors in a response while testing today. In a post, I have this content:

```html
<!-- wp:paragraph -->
<p>Unnested paragraph.</p>
<!-- /wp:paragraph -->

<!-- wp:columns -->
<div class="wp-block-columns"><!-- wp:column -->
<div class="wp-block-column"><!-- wp:paragraph -->
<p>Column 1</p>
<!-- /wp:paragraph --></div>
<!-- /wp:column -->

<!-- wp:column -->
<div class="wp-block-column"><!-- wp:paragraph -->
<p>Column 2</p>
<!-- /wp:paragraph --></div>
<!-- /wp:column -->

<!-- wp:column -->
<div class="wp-block-column"><!-- wp:paragraph -->
<p>Column 3</p>
<!-- /wp:paragraph --></div>
<!-- /wp:column --></div>
<!-- /wp:columns -->
```

---

image

---

And then I query for the blocks/innerBlocks of that post:

```gql
query NewQuery {
  post(id: 419, idType: DATABASE_ID) {
    blocksData {
      blocks {
        attributes {
          name
          value
        }
        id
        name
        innerBlocks {
          attributes {
            name
            value
          }
          id
          name
          parentId
        }
      }
    }
  }
}
```

The result appears to hold the correct blocks, but also a bunch of `Internal server error` in an `errors` array:

<details>

<summary>Click to expand</summary>

```json
{
  "errors": [
    {
      "message": "Internal server error",
      "extensions": {
        "category": "internal"
      },
      "locations": [
        {
          "line": 12,
          "column": 11
        }
      ],
      "path": [
        "post",
        "blocksData",
        "blocks",
        1,
        "innerBlocks",
        0,
        "attributes"
      ]
    },
    {
      "message": "Internal server error",
      "extensions": {
        "category": "internal"
      },
      "locations": [
        {
          "line": 12,
          "column": 11
        }
      ],
      "path": [
        "post",
        "blocksData",
        "blocks",
        1,
        "innerBlocks",
        2,
        "attributes"
      ]
    },
    {
      "message": "Internal server error",
      "extensions": {
        "category": "internal"
      },
      "locations": [
        {
          "line": 12,
          "column": 11
        }
      ],
      "path": [
        "post",
        "blocksData",
        "blocks",
        1,
        "innerBlocks",
        4,
        "attributes"
      ]
    }
  ],
  "data": {
    "post": {
      "blocksData": {
        "blocks": [
          {
            "attributes": [
              {
                "name": "content",
                "value": "Unnested paragraph."
              },
              {
                "name": "dropCap",
                "value": ""
              }
            ],
            "id": "1",
            "name": "core/paragraph",
            "innerBlocks": null
          },
          {
            "attributes": [
              {
                "name": "isStackedOnMobile",
                "value": "1"
              }
            ],
            "id": "2",
            "name": "core/columns",
            "innerBlocks": [
              {
                "attributes": null,
                "id": "3",
                "name": "core/column",
                "parentId": "2"
              },
              {
                "attributes": [
                  {
                    "name": "content",
                    "value": "Column 1"
                  },
                  {
                    "name": "dropCap",
                    "value": ""
                  }
                ],
                "id": "4",
                "name": "core/paragraph",
                "parentId": "3"
              },
              {
                "attributes": null,
                "id": "5",
                "name": "core/column",
                "parentId": "2"
              },
              {
                "attributes": [
                  {
                    "name": "content",
                    "value": "Column 2"
                  },
                  {
                    "name": "dropCap",
                    "value": ""
                  }
                ],
                "id": "6",
                "name": "core/paragraph",
                "parentId": "5"
              },
              {
                "attributes": null,
                "id": "7",
                "name": "core/column",
                "parentId": "2"
              },
              {
                "attributes": [
                  {
                    "name": "content",
                    "value": "Column 3"
                  },
                  {
                    "name": "dropCap",
                    "value": ""
                  }
                ],
                "id": "8",
                "name": "core/paragraph",
                "parentId": "7"
              }
            ]
          }
        ]
      }
    }
  },
  "extensions": {
    "debug": [
      {
        "type": "DEBUG_LOGS_INACTIVE",
        "message": "GraphQL Debug logging is not active. To see debug logs, GRAPHQL_DEBUG must be enabled."
      }
    ]
  }
}
```

</details>

ingeniumed:
@alecgeatches So the error itself was due to the fact that we were setting the attributes to an empty object array whenever it was empty. [This](https://stackoverflow.com/a/64988972) answer has a good explanation for this, and recommends using `ArrayObject` instead. I switched that over, and have added an extra check for the attributes to ensure our graphQL code doesn't run if its empty. That should resolve it.

tests/graphql/test-graphql-api.php
```diff
@@ -0,0 +1,135 @@
...
+	public function test_is_graphql_enabled_false() {
+		$is_graphql_enabled_function = function () {
+			return false;
+		};
+		add_filter( 'vip_block_data_api__is_graphql_enabled', $is_graphql_enabled_function, 10, 0 );
+		$this->assertFalse( apply_filters( 'vip_block_data_api__is_graphql_enabled', true ) );
+		remove_filter( 'vip_block_data_api__is_graphql_enabled', $is_graphql_enabled_function, 10, 0 );
+	}
+
+	public function test_get_blocks_data() {
```
ingeniumed, line 28: This one test battle tests the entire graphql code. We can test the exact GraphQL query as we need wp-graphQL so this is a good middle ground for that.

README.md
smithjw1, line 1: Do we need to add documentation here if they want to use GraphQL?

* How to turn it on
* What other plugins are necessary
* What the output looks like

src/graphql/graphql-api.php
smithjw1: If it is on by default, and a site has WP GraphQL enabled what happens to the existing JSON endpoint? Does it still return JSON?

README.md
alecgeatches: A documentation section on WPGraphQL integration makes sense! First, so users know it's supported, and I also think some code snippets for these would be useful:

- An example WPGraphQL query
- An example response
- Maybe a small amount of Node.js code that can be used to reconstruct a block hierarchy

None of these need to be huge examples, I think just a nested block or two as an example. Maybe a `core/list` with two items?

src/graphql/graphql-api.php
ingeniumed: Yip, the existing endpoint will work just as it previously did. While testing this, I used the regular REST endpoint in tandem as well.

README.md
ingeniumed: Great point, and something I neglected while trying to move this out of draft status.

ingeniumed: I have revamped the README to add in the graphQL details, so check it out. What's missing is the node utility to reconstruct the block hierarchy, which I am working on. I'll have that by monday.

ingeniumed:
One thing I did btw - I moved the analytics usage code out of the rest API and into the parser itself. This way in case someone tries to call the method directly, instead of going through the rest api they'll still be counted. The same applies for the graphQL code as well.

Happy to revert this, but figured this was a good time to rectify that.

README.md
```diff
@@ -6,35 +6,32 @@
   <img src="https://github.com/Automattic/vip-block-data-api/blob/media/vip-block-data-api-animation-830.gif" alt="VIP Block Data API attribute sourcing animation" />
 </picture>
 
-The Block Data API is a REST API for retrieving block editor posts structured as JSON data. While primarily designed for use in decoupled WordPress, the Block Data API can be used anywhere you want to represent block markup as structured data.
+The Block Data API is an API for retrieving block editor posts structured as JSON data. This API is available as a REST, as well as a GraphQL API. While primarily designed for use in decoupled WordPress, the Block Data API can be used anywhere you want to represent block markup as structured data.
```
chriszarate:
```suggestion
The Block Data API is an API for retrieving block editor posts structured as JSON data, with integrations for both the official WordPress REST API and WPGraphQL. While primarily designed for use in decoupled WordPress, the Block Data API can be used anywhere you want to represent block markup as structured data.
```

README.md
ingeniumed: The utility function has been added as well! I tried to simplify it further, but ran into some recursion problems so decided to leave it as it is. This can be updated further in an another update if we want.

smithjw1 (approved):
I opened a PR with some wording changes to the README. Looks great!

src/graphql/graphql-api.php
```diff
@@ -0,0 +1,245 @@
...
+defined( 'ABSPATH' ) || die();
+
+/**
+ * GraphQL API to offer an alternative to the REST API.
+ */
+class GraphQLApi {
+	/**
+	 * Initiatilize the graphQL API, if its allowed
+	 *
+	 * @access private
```
chriszarate: not private

src/graphql/graphql-api.php
```diff
@@ -0,0 +1,245 @@
...
+namespace WPCOMVIP\BlockDataApi;
+
+defined( 'ABSPATH' ) || die();
+
+/**
+ * GraphQL API to offer an alternative to the REST API.
+ */
+class GraphQLApi {
+	/**
+	 * Initiatilize the graphQL API, if its allowed
```
chriszarate: This comment could be better (either here or below), to explain that it hooks into the `graphql_register_types` action, which only fires if WPGraphQL is installed and enabled, and is further controlled by the `vip_block_data_api__is_graphql_enabled` filter.

src/graphql/graphql-api.php
```diff
@@ -0,0 +1,245 @@
...
+	 *
+	 * @access private
+	 */
+	public static function init() {
+		/**
+		 * Filter to enable/disable the graphQL API. By default, it is enabled.
+		 * 
+		 * @param bool $is_graphql_to_be_enabled Whether the graphQL API should be enabled or not.
+		 */
+		$is_graphql_to_be_enabled = apply_filters( 'vip_block_data_api__is_graphql_enabled', true );
```
chriszarate: This `init` method is being called on load (at the end of this file, not hooked into WordPress `init` or another lifecycle hook). Therefore, plugin / theme code may not be able to add a filter before this is executed. I would move this inside the `register_types` method.

src/graphql/graphql-api.php
```diff
@@ -0,0 +1,245 @@
...
+
+		if ( ! $is_graphql_to_be_enabled ) {
+			return;
+		}
+
+		add_filter( 'vip_block_data_api__sourced_block_result', [ __CLASS__, 'transform_block_format' ], 10, 5 );
```
chriszarate: Do you want to add this filter if WPGraphQL isn't installed or active?

src/graphql/graphql-api.php
```diff
@@ -0,0 +1,245 @@
...
+		add_action( 'graphql_register_types', [ __CLASS__, 'register_types' ] );
+	}
+
+	/**
+	 * Extract the blocks data for a post, and return back in the format expected by the graphQL API.
+	 *
+	 * @param  WPGraphQL\Model\Post $post_model Post model for post.
+	 * @return array
+	 */
+	public static function get_blocks_data( $post_model ) {
```
chriszarate, line 34: Is there a PHP version policy for this plugin? Can you use type declarations?
```suggestion
	public static function get_blocks_data( \WPGraphQL\Model\Post $post_model ) {
```

src/graphql/graphql-api.php
chriszarate: What kind of problem?

src/graphql/graphql-api.php
```diff
@@ -0,0 +1,245 @@
...
+		$post_id        = $post_model->ID;
+		$post           = get_post( $post_id );
+		$filter_options = [ 'graphQL' => true ];
+
+		$content_parser = new ContentParser();
+
+		$parser_results = $content_parser->parse( $post->post_content, $post_id, $filter_options );
```
chriszarate: can access on $model->fields
```suggestion
		$parser_results = $content_parser->parse( $post_model->fields->contentRaw, $post_model->fields->ID, $filter_options );
```

src/graphql/graphql-api.php
```diff
@@ -0,0 +1,245 @@
...
+	 *
+	 * @param array  $sourced_block An associative array of parsed block data with keys 'name' and 'attribute'.
+	 * @param string $block_name Name of the parsed block, e.g. 'core/paragraph'.
+	 * @param int    $post_id Post ID associated with the parsed block.
+	 * @param array  $block Result of parse_blocks() for this block. Contains 'blockName', 'attrs', 'innerHTML', and 'innerBlocks' keys.
+	 * @param array  $filter_options Options to filter using, if any.
+	 *
+	 * @return array
+	 */
+	public static function transform_block_format( $sourced_block, $block_name, $post_id, $block, $filter_options ) { // phpcs:disable Generic.CodeAnalysis.UnusedFunctionParameter.FoundAfterLastUsed
```
chriszarate: I don't understand why this is called via a filter from `ContentParser`. Why not just call it directly in `get_blocks_data`?

src/graphql/graphql-api.php
```diff
@@ -0,0 +1,245 @@
...
+
+	/**
+	 * Register types and fields graphql integration.
+	 *
+	 * @return void
+	 */
+	public static function register_types() {
+		// Register the type corresponding to the attributes of each individual block.
+		register_graphql_object_type(
+			'BlockDataAttribute',
```
chriszarate: Naming is hard, but isn't this a `BlockAttribute`? BlockData is the collection?

src/graphql/graphql-api.php
```diff
@@ -0,0 +1,245 @@
...
+			[
+				'description' => 'Block data attribute',
+				'fields'      => [
+					'name'  => [
+						'type'        => 'String',
```
chriszarate: It's good practice to designate fields as non-nullable whenever possible. That way the consumer can have confidence that a field will always have a value and can skip inspection.
```suggestion
						'type'        => [ 'non_null' => 'String' ],
```

src/graphql/graphql-api.php
```diff
@@ -0,0 +1,245 @@
...
+		);
+
+		// Register the type corresponding to the individual inner block, with the above attribute.
+		register_graphql_type(
+			'InnerBlockData',
+			[
+				'description' => 'Block data',
+				'fields'      => [
+					'id'         => [
+						'type'        => 'String',
```
chriszarate: IDs in WPGraphQL are always non-nullable and are typically of type `ID` meaning a Relay ID. [Example](https://github.com/wp-graphql/wp-graphql/blob/develop/src/Type/InterfaceType/Node.php#L20). Create a Relay ID with `Relay::toGlobalId` [example](https://github.com/wp-graphql/wp-graphql/blob/1051357687339e53b7bd19b816c756316dbc8a47/src/Model/Menu.php#L72).
```suggestion
						'type'        => [ 'non_null' => 'ID' ],
```

src/graphql/graphql-api.php
```diff
@@ -0,0 +1,245 @@
...
+						'description' => 'ID of the block',
+					],
+					'parentId'   => [
+						'type'        => 'String',
```
chriszarate: review all these fields for non-nullability

src/graphql/graphql-api.php
```diff
@@ -0,0 +1,245 @@
...
+		);
+
+		// Register the type corresponding to the individual block, with the above attribute.
+		register_graphql_type(
+			'BlockData',
+			[
+				'description' => 'Block data',
+				'fields'      => [
+					'id'          => [
+						'type'        => 'String',
```
chriszarate:
```suggestion
						'type'        => [ 'non_null' => ID ],
```

src/graphql/graphql-api.php
```diff
@@ -0,0 +1,245 @@
...
+						],
+						'description' => 'Block data attributes',
+					],
+				],
+			],
+		);
+
+		// Register the type corresponding to the individual block, with the above attribute.
+		register_graphql_type(
+			'BlockData',
```
chriszarate, line 176: It seems like you don't need two distinct types for Blocks and InnerBlocks. You could have the same type with nullable fields for `parentId` and `innerBlocks`?

src/graphql/graphql-api.php
```diff
@@ -0,0 +1,245 @@
...
+					],
+					'attributes'  => [
+						'type'        => [
+							'list_of' => 'BlockDataAttribute',
+						],
+						'description' => 'Block data attributes',
+					],
+					'innerBlocks' => [
+						'type'        => [ 'list_of' => 'InnerBlockData' ],
+						'description' => 'Flattened list of inner blocks of this block',
```
chriszarate, line 200: I get why these are flattened, but is the "reparenting" going to be annoying / error-prone? Is the nesting issue severe enough to warrant this approach?

src/graphql/graphql-api.php
```diff
@@ -0,0 +1,245 @@
...
+		register_graphql_type(
+			'BlocksData',
+			[
+				'description' => 'Data for all the blocks',
+				'fields'      => [
+					'blocks'   => [
+						'type'        => [ 'list_of' => 'BlockData' ],
+						'description' => 'List of blocks data',
+					],
+					'warnings' => [
```
chriszarate, line 236: Warning / error data typically doesn't ship in the `data` payload, but instead [goes to `graphql_debug()`](https://www.wpgraphql.com/docs/debugging#graphql_debug) where it can be conditionally displayed in the `errors` payload. Better control for the site maintainer.

src/graphql/graphql-api.php
```diff
@@ -0,0 +1,245 @@
...
+						'type'        => [ 'list_of' => 'String' ],
+						'description' => 'List of warnings related to processing the blocks data',
+					],
+				],
+			],
+		);
+
+		// Register the field on every post type that supports 'editor'.
+		register_graphql_field(
+			'NodeWithContentEditor',
```
chriszarate, line 246: I know you are following the pattern of the decoupled bundle here, but let's discuss the pros and cons of adding the field to this union.

src/parser/block-additions/core-image.php
```diff
@@ -29,12 +29,13 @@ public static function init() {
...
 	 * @param int|null $post_id Id of the post.
 	 * @param array    $block Block being parsed.
+	 * @param array    $filter_options Options for the filter, if any.
 	 * 
 	 * @access private
 	 *
 	 * @return array Updated sourced block with new metadata information
 	 */
-	public static function add_image_metadata( $sourced_block, $block_name, $post_id, $block ) { // phpcs:disable Generic.CodeAnalysis.UnusedFunctionParameter.FoundAfterLastUsed
+	public static function add_image_metadata( $sourced_block, $block_name, $post_id, $block, $filter_options ) { // phpcs:disable Generic.CodeAnalysis.UnusedFunctionParameter.FoundAfterLastUsed
```
chriszarate: unused?

src/parser/block-additions/core-image.php
```diff
@@ -19,7 +19,7 @@ class CoreImage {
 	 * @access private
 	 */
 	public static function init() {
-		add_filter( 'vip_block_data_api__sourced_block_result', [ __CLASS__, 'add_image_metadata' ], 5, 4 );
+		add_filter( 'vip_block_data_api__sourced_block_result', [ __CLASS__, 'add_image_metadata' ], 5, 5 );
```
chriszarate: seems like you could just keep the arity at 4

src/rest/rest-api.php
```diff
@@ -128,8 +128,6 @@ public static function get_block_content( $params ) {
 		$post_id = $params['id'];
 		$post    = get_post( $post_id );
 
-		Analytics::record_usage();
```
chriszarate, line 131: ?

ingeniumed: Due to the way we were recording usage, if someone decided to call the `ContentParser` directly instead of going via the rest api the usage would not be counted. By moving the usage tracker into the `ContentParser` it avoids that problem. Plus it works for GraphQL that way as well, since that calls that code as well.

src/graphql/graphql-api.php
ingeniumed: The core problem we wanted to solve was that, you should be able to get back the complicated block hierarchy in a post without having any knowledge of the depth involved. This gets us past that, and to ensure that the re-parenting isn't annoying we added a little utility function in the README that could be used. Admittedly that function could be simplified further, but I kept it simple enough to follow. By ensuring that each block gets an id and if applicable, a parent id the likelihood of an error should be minimal (one would hope,  but obviously things can go wrong).

The other approach would be to flatten the innerBlocks alongside all the blocks but that proved to be even more complicated (in terms of code and the outcome) and wouldn't necessarily match the structure of a parent-child in a post with Gutenberg blocks.

So with that in mind, this was the ultimate solution that was picked to allow graphQL to be added on top of the block data api without the innerBlocks being a problem.

src/graphql/graphql-api.php
ingeniumed: The reason I had kept it separate was because the outer blocks will have only `innerBlocks` but the innerBlocks will only `parentId`.  I didn't want to potentially allow a query where you could build `innerBlocks` at any depth. This way you are restricted to what we want, while still having some flexibility like you would in GraphQL land.

Were you suggesting that by doing it the way you mentioned, this would not be a concern? Hope I understood that right.

src/graphql/graphql-api.php
```diff
@@ -0,0 +1,245 @@
...
+			'BlocksData',
```
chriszarate: field name should be camel case
```suggestion
			'blocksData',
```

src/graphql/graphql-api.php
chriszarate: I basically just saying use the same type for both inner and outer blocks and use `null` appropriately, like:

```json
{
   "blockData": [
      {
         "id": "abc123=",
         "parentId": null,
         "name": "core/some-outer-block",
         "attributes": [],
         "innerBlocks": [
            {
               "id": "def456=",
               "parentId": "abc123=",
               "name": "core/some-outer-block",
               "attributes": [],
               "innerBlocks": null
            }
         ]
      }
   ]
}
```

src/graphql/graphql-api.php
chriszarate: Ok. I totally get the why, just wondering how big a deal this actually is in practice:

> you should be able to get back the complicated block hierarchy in a post without having any knowledge of the depth involved

If you use fragments in your query, adding [another layer isn't a huge deal](https://github.com/Automattic/vip-go-nextjs-skeleton/blob/b6ad3127e51c5d032255488665956ef1cdd58b9d/graphql/fragments/ContentNode.graphql#L9-L12). And since you are reconstructing the deep tree on the client side, your block parsing code isn't less complex.

Yes, if you add another level beyond what you are querying for, it simply won't be in the response. But adding a `hasInnerBlocks` or `innerBlocksCount` property could allow the queryer to determine if they are "missing out" on blocks.

Another drawback of this approach is that it makes human inspection of the response data much less friendly, possibly impossible. Human debugging is useful and common, especially with the built-in GraphiQL client.

Again, totally understand where this is coming from, just want to make sure the devex has been fully considered.

chriszarate: Just an idea: If you use the same type for both inner and outer blocks, you could add a [GraphQL field directive](https://graphql.org/learn/queries/) that allows the queryer to choose whether to flatten or not. If would look like this:

```gql
query FlattenBlocks {
   post(id: "abc123=") {
      blockData(flatten: true) {
         # ...
      }
   }
}

query NestedBlocks {
   post(id: "abc123=") {
      blockData(flatten: false) { # or omit the directive
         # ...
      }
   }
}
```

smithjw1: In the research specifically mentioned flattening the hierarchy and using parent/child relationships. No one balked at it.

I'm not trying to suggest that flattening is the right choice. I think optionality is good, as are Chris' points.

My take is we have options that customers will accept, and we can choose from among them, which is a great place to be.

src/graphql/graphql-api.php
ingeniumed: There's 2 reasons why it was done that way:

- I wanted to keep the GraphQL logic separate from the parsing logic
- I also wanted to ensure that, the transforming happened at the same time as when the parser worked so we didn't loop over each block twice

So this way, while its odd, gives us the best of both worlds. With that said, I do think it can be done in a better way even if that means we go over the entire block list twice.

Talking it over with @alecgeatches, I'll change it up to:

- Get the parser results from ContentParser
- Go over the results, and transform it into the graphQL format
- This would make any potential changes to the structure, like flattening or not flattening innerBlocks easier as well as anything else we think of.
- No more filter oddeties to be done

src/graphql/graphql-api.php
ingeniumed: Yip, I understand what you mean. @alecgeatches had suggested something similar and I had resisted doing it but you are both right on this. It would make it simpler overall, so I'll make that change.

src/graphql/graphql-api.php
ingeniumed: I had looked up [this](https://github.com/wp-graphql/wp-graphql/blob/develop/src/Type/InterfaceType/NodeWithContentEditor.php) page as well as [this](https://woographql.com/schema/definition/-node-with-content-editor) one to see what it could support. From what I gathered this would be what we support - pages and posts that would be made using the content editor. We have a check with the contentParser that already checks to see if the provided post has blocks within it or not, so that should account for the classic editor.

That was my thinking, alongside this being used in the decoupled bundle already. Was there something else or a downside that I might have missed in using this?

src/graphql/graphql-api.php
ingeniumed: I have implemented this change now, so it's a shared type with the nullable set for parentId and innerBlocks. In case I haven't done the nullable part right, lemme know.

src/graphql/graphql-api.php
ingeniumed: I'm still working on this, ran into some problems that are of my own making.

src/graphql/graphql-api.php
ingeniumed: We have set it at 7.4, but will be upping it to 8.0. At that point we were gonna go in and set the type declarations. That is on the cards, so I haven't done that here.

That said, I do have to figure out when I do how to pass in post model of type \WPGraphQL\Model\Post when we dont bundle in WP-GraphQL with the plugins. The tests fail due to it, I realized.

I'll leave this one as is for this PR

src/graphql/graphql-api.php
ingeniumed: I had tried to do it that way, but no matter what I did the fields were null. That's why I did it this way. The plugin internally calls get_post already so the info is quickly returned.

Is there anyone special I need to do? I had called it the way u suggested 🤔

src/graphql/graphql-api.php
ingeniumed: Switched it over so the transformation is done after the parser comes back, so more filter weirdness

src/graphql/graphql-api.php
ingeniumed: Yip, changed it

src/graphql/graphql-api.php
ingeniumed: Good point, I have done this now

src/graphql/graphql-api.php
ingeniumed: Good point, I have done this now for the ID and the parentId.

src/graphql/graphql-api.php
ingeniumed: Good point, I have done this now for all the fields

src/graphql/graphql-api.php
ingeniumed: Good point, I have done this now

src/parser/block-additions/core-image.php
ingeniumed: Yip, cleaned this up

src/parser/block-additions/core-image.php
ingeniumed: Made this change

src/graphql/graphql-api.php
ingeniumed: Changed this as well

ingeniumed:
The test I added is now failing because Relay cannot be found since wp_graphQL is not bundled with the plugin when running the tests. Gonna try and fix this so the test can stay

src/graphql/graphql-api.php
ingeniumed: I realized I summed up the same comment that's in the `blocks.php` within the decoupled bundle itself. That's really why it's been done this way as that same reasoning applies.


```
	// WPGraphQL's Post model restricts access to the raw post_content (contentRaw)
	// based on "edit_posts" cap. Since we want to serve blocks even to logged-out
	// users -- and because we are parsing this content before returning it --
	// we'll bypass the model and access the raw content directly.
```

src/graphql/graphql-api.php
ingeniumed: Thanks to the re-write I did of the way the parser's result is transformed into the right graphQL schema, I was able to add in this flatten option. So by default it's set to true because we want the inner blocks to be flattened. Omitting it sets it to true by default. If you set it to false, it will not flatten the inner blocks, and send back the original hierarchy.

So best of both options are available, but we give back the flattened hierarchy by default as thats the best option we want to be used.

src/graphql/graphql-api.php
ingeniumed: I have taken it out now entirely.

src/graphql/graphql-api.php
ingeniumed: Good point, I have moved this into the `register_types` now.

src/graphql/graphql-api.php
ingeniumed: I have made this change as well

src/graphql/graphql-api.php
ingeniumed: Taken it out

src/graphql/graphql-api.php
ingeniumed: Screenshot 2023-12-08 at 17 20 31

So if I return $parser_results as is, the above error is what I get. Maybe it should not be returned this way and I'm going about it incorrectly?

ingeniumed: Screenshot 2023-12-08 at 17 22 30

vs the way I did it right now

ingeniumed:
I have made all the changes that are necessary, and there are 2 pending questions that I commented on. So this should be good to review once more

README.md
```diff
@@ -281,7 +287,211 @@ Examples of WordPress block markup and the associated data structure returned by
...
 </table>
 
-## Preact example
+### GraphQL
+
+The GraphQL API, requires some setup before it can be it can be used.
+
+#### Setup
+
+The Block Data API integrates with **WP-GraphQL** to provide a GraphQL API as well. As a result, it is necessary to have WP-GraphQL installed, to activate the GraphQL API. To do so, follow the instructions mentioned [here](https://www.wpgraphql.com/docs/quick-start#install).
```
chriszarate: There isn't a separate step to activate the GraphQL endpoint, it's just plugin activation. Also linking the word "here" is a pet peeve (and bad for SEO). :)

```suggestion
The Block Data API integrates with **WPGraphQL** to provide a GraphQL API. It is necessary to have WPGraphQL [installed and activated](https://www.wpgraphql.com/docs/quick-start#install).
```

README.md
```diff
@@ -281,7 +287,211 @@ Examples of WordPress block markup and the associated data structure returned by
...
-## Preact example
+### GraphQL
+
+The GraphQL API, requires some setup before it can be it can be used.
+
+#### Setup
+
+The Block Data API integrates with **WP-GraphQL** to provide a GraphQL API as well. As a result, it is necessary to have WP-GraphQL installed, to activate the GraphQL API. To do so, follow the instructions mentioned [here](https://www.wpgraphql.com/docs/quick-start#install). 
+
+Once WP-GraphQL has been installed and setup, a new field called `blocksData` will show up under categories that support `ContentNodes` like posts, pages, etc.
```
chriszarate:
```suggestion
Once WPGraphQL has been installed and setup, a new field called `blocksData` will be available for post types that provide content, like posts, pages, etc. 
```

README.md
```diff
@@ -281,7 +287,211 @@ Examples of WordPress block markup and the associated data structure returned by
 </tr>
 </table>
 
-## Preact example
+### GraphQL
+
+The GraphQL API, requires some setup before it can be it can be used.
```
chriszarate:
```suggestion
The GraphQL API requires some setup before it can be it can be used.
```

README.md
```diff
@@ -281,7 +287,211 @@ Examples of WordPress block markup and the associated data structure returned by
...
+
+The Block Data API integrates with **WP-GraphQL** to provide a GraphQL API as well. As a result, it is necessary to have WP-GraphQL installed, to activate the GraphQL API. To do so, follow the instructions mentioned [here](https://www.wpgraphql.com/docs/quick-start#install). 
+
+Once WP-GraphQL has been installed and setup, a new field called `blocksData` will show up under categories that support `ContentNodes` like posts, pages, etc. 
+
+#### Usage
+
+The `blocksData` field would be the field through which block data would be returned under a category that supports it. 
+
+An example of what a query would look like for a post:
```
chriszarate: Category has a different meaning in a WP context.
```suggestion
The `blocksData` field provides block data for post types that support it. Here is an example query:
```

README.md
```diff
@@ -281,7 +287,211 @@ Examples of WordPress block markup and the associated data structure returned by
...
+            value
+          }
+        }
+      }
+    }
+  }
+}
+```
+
+Here, the `id` and `parentId` fields are dynamically generated, unique IDs that help to identify parent-child relationships in the `innerBlocks` under a block, in the overall block structure.The resulting `innerBlocks` is a flattened list, that can be untangled using the combination of `id` and `parentId` fields. This is helpful in being able to give back a complicated nesting structure, without having any knowledge as to how deep this nesting goes.
```
chriszarate:
```suggestion
Here, the `id` and `parentId` fields are dynamically generated, unique IDs that help to identify parent-child relationships in the `innerBlocks` under a block, in the overall block structure. The resulting `innerBlocks` is a flattened list, that can be untangled using the combination of `id` and `parentId` fields. This is helpful in being able to give back a complicated nesting structure, without having any knowledge as to how deep this nesting goes. 
```

README.md
```diff
@@ -281,7 +287,211 @@ Examples of WordPress block markup and the associated data structure returned by
...
+        }
+      }
+    }
+  }
+}
+```
+
+Here, the `id` and `parentId` fields are dynamically generated, unique IDs that help to identify parent-child relationships in the `innerBlocks` under a block, in the overall block structure.The resulting `innerBlocks` is a flattened list, that can be untangled using the combination of `id` and `parentId` fields. This is helpful in being able to give back a complicated nesting structure, without having any knowledge as to how deep this nesting goes. 
+
+In addition, the attributes of a block are a list of `name`, `value` pairs so as to avoid having to figure out the type of each block. This allows giving back any block's attributes.
```
chriszarate: To be fair here, it doesn't avoid having to figure out the type. The consumer is likely going to need to do exactly that and perform type conversion. You might want to discuss the trade-offs of this decision here.

```suggestion
In addition, the attributes of a block are a list of `name` / `value` pairs so as to avoid having to figure out the type of each block. This allows giving back any block's attributes.
```

README.md
```diff
@@ -281,7 +287,211 @@ Examples of WordPress block markup and the associated data structure returned by
...
+                          {
+                              "name": "ordered",
+                              "value": ""
+                          },
+                          {
+                              "name": "values",
+                              "value": ""
+                          }
+                      ],
+                      "id": "1",
```
chriszarate: Use example Relay IDs?

README.md
```diff
@@ -381,6 +591,65 @@ The code above produces this HTML from post data:
 </div>
 ```
 
+### Utility function to reconstruct the block hierarchy
+
+The purpose of this function is to take the flattened `innerBlock` list under each root block, and reconstruct the block hierarchy.
```
chriszarate:
```suggestion
The purpose of this function is to take the flattened `innerBlocks` list under each root block, and reconstruct the block hierarchy.
```

README.md
```diff
@@ -381,6 +591,65 @@ The code above produces this HTML from post data:
...
+1. Go over each block given back.
+2. Go over each block's `innerBlocks`.
+   * For each `innerBlock`, check if the `parentId` matches the `id` of the root block.
+   * If yes, add that `innerBlock` to a new list.
+   * If no, go over the newly constructed list and repeat step 2's logic as the block could be nested under another `innerBlock`.
+
+This logic has been split over two functions, with the core logic (steps 1, 2a, 2b) being in the function below and the recursive case (2c) being handled in the second function called `convertInnerBlocksToHierarchy`.
+
+```js
+const blocks = payload.data?.post?.blocksData?.blocks;
```
chriszarate: blocks is possibly undefined
```suggestion
const blocks = payload.data?.post?.blocksData?.blocks ?? [];
```

README.md
```diff
@@ -754,7 +1023,28 @@ Note that custom block filter rules can also be created in code via [the `vip_bl
...
+
+```php
+/**
+	* Filter to enable/disable the graphQL API. By default, it is enabled.
+	* 
+	* @param bool $is_graphql_to_be_enabled Whether the graphQL API should be enabled or not.
+*/
+apply_filters( 'vip_block_data_api__is_graphql_enabled', true );
+```
+
```
chriszarate: It's a little odd to include the `apply_filters` definition here, which isn't relevant to usage. I would remove it

```suggestion
```

README.md
```diff
@@ -754,7 +1023,28 @@ Note that custom block filter rules can also be created in code via [the `vip_bl
...
+To disable the GraphQL API, the following can be done:
+
+```php
+add_filter( 'vip_block_data_api__is_graphql_enabled', function( ) {
+    return false;
+}, 10, 1);
```
chriszarate:
```suggestion
add_filter( 'vip_block_data_api__is_graphql_enabled', '__return_false' );
```

src/graphql/graphql-api.php
chriszarate: Code is fine as-is, was just curious

src/graphql/graphql-api.php
```diff
@@ -0,0 +1,255 @@
...
+
+		// We need to not return a WP_Error object, and so a regular exception is returned.
+		if ( is_wp_error( $parser_results ) ) {
+			Analytics::record_error( $parser_results );
+
+			// Return API-safe error with extra data (e.g. stack trace) removed.
+			return new \Exception( $parser_results->get_error_message() );
+		}
+
+		$parser_results['blocks'] = array_map(function ( $block ) {
```
chriszarate: is there no phpcs on this repo?
```suggestion
		$parser_results['blocks'] = array_map( function ( $block ) {
```

src/graphql/graphql-api.php
```diff
@@ -0,0 +1,255 @@
...
+			return self::transform_block_format( $block );
+		}, $parser_results['blocks'] );
+
+		return $parser_results;
+	}
+
+	/**
+	 * Transform the block's format to the format expected by the graphQL API.
+	 *
+	 * @param array $block An associative array of parsed block data with keys 'name' and 'attribute'.
```
chriszarate: +1 for type declarations instead of hand-written descriptions
```suggestion
	 * @param array $block An associative array of parsed block data with keys 'name' and 'attributes'.
```

src/graphql/graphql-api.php
```diff
@@ -0,0 +1,255 @@
...
+			}, $block['innerBlocks'] );
+		}
+
+		return $block;
+	}
+
+	/**
+	 * Convert the attributes to be in the name-value format that the schema expects.
+	 *
+	 * @param array $block An associative array of parsed block data with keys 'name' and 'attribute'.
```
chriszarate:
```suggestion
	 * @param array $block An associative array of parsed block data with keys 'name' and 'attributes'.
```

src/graphql/graphql-api.php
```diff
@@ -0,0 +1,255 @@
...
+	 *
+	 * @return array
+	 */
+	public static function map_attributes( $block ) {
+		// check if type of attributes is stdClass and unset it as that's not supported by graphQL.
+		if ( isset( $block['attributes'] ) && is_object( $block['attributes'] ) ) {
+			unset( $block['attributes'] );
```
chriszarate, line 91: how / why would this happen?

src/graphql/graphql-api.php
```diff
@@ -0,0 +1,255 @@
...
+		} elseif ( isset( $block['attributes'] ) && ! empty( $block['attributes'] ) ) {
+			$block['attributes'] = array_map(
+				function ( $name, $value ) {
+					return [
+						'name'  => $name,
+						'value' => $value,
```
chriszarate: I guess WPGraphQL is doing the casting to string? You might want to do it here in code just to be explicit and test it.

src/graphql/graphql-api.php
```diff
@@ -0,0 +1,255 @@
...
+	 * @param array  $flattened_blocks the flattened blocks that's built up as we go through the inner blocks.
+	 * 
+	 * @return array
+	 */
+	public static function flatten_inner_blocks( $inner_blocks, $parent_id, $flattened_blocks = [] ) {
+		foreach ( $inner_blocks as $inner_block ) {
+				// Set the parentId to be the ID of the parent block whose inner blocks are being flattened.
+				$inner_block['parentId'] = $parent_id;
+
+				// This block doesnt have any inner blocks, so just add it to the flattened blocks. Ensure the parentId is set.
```
chriszarate: indenting is off here

src/graphql/graphql-api.php
```diff
@@ -0,0 +1,255 @@
...
+			[
+				'description' => 'Block data',
+				'fields'      => [
+					'id'          => [
+						'type'        => [ 'non_null' => 'ID' ],
+						'description' => 'ID of the block',
+					],
+					'parentId'    => [
+						'type'        => 'ID',
+						'description' => 'ID of the parent for this inner block, if it is an inner block. Otherwise, it will not be set. If it set, this will match the ID of the block',
```
chriszarate: last part confusing / redundant
```suggestion
						'description' => 'ID of the parent for this inner block, if it is an inner block. Otherwise, it will be null.',
```

src/graphql/graphql-api.php
chriszarate: Nice! FYI it is not mentioned in the README

src/graphql/graphql-api.php
```diff
@@ -0,0 +1,255 @@
...
+			[
+				'description' => 'Data for all the blocks',
+				'fields'      => [
+					'blocks'   => [
+						'type'        => [ 'list_of' => 'BlockData' ],
+						'description' => 'List of blocks data',
+						'args'        => [
+							'flatten' => [
+								'type'        => 'Boolean',
+								'description' => 'Collate the inner blocks under each root block into a single list with a parent-child relationship. This is set to true by default, and setting it to false will reverse that to preserve the original block hierarchy, at the expense of knowing the exact depth when querying the inner blocks. Default: true',
```
chriszarate:
```suggestion
								'description' => 'Collate the inner blocks under each root block into a single list with a parent-child relationship. This is set to true by default, and setting it to false will preserve the original block hierarchy, but will require nested inner block queries to the desired depth. Default: true',
```

src/graphql/graphql-api.php
```diff
@@ -0,0 +1,255 @@
...
+							],
+						],
+						'resolve'     => function ( $blocks, $args ) {
+							if ( ! isset( $args['flatten'] ) || true === $args['unflatten'] ) {
```
chriszarate: bug, probably needs tests
```suggestion
							if ( ! isset( $args['flatten'] ) || false === $args['flatten'] ) {
```

src/graphql/graphql-api.php
```diff
@@ -0,0 +1,255 @@
...
+								$blocks['blocks'] = array_map(function ( $block ) {
+									// Flatten the inner blocks, if any.
+									if ( isset( $block['innerBlocks'] ) ) {
+										$block['innerBlocks'] = self::flatten_inner_blocks( $block['innerBlocks'], $block['id'] );
+									}
+
+									return $block;
+								}, $blocks['blocks'] );
```
chriszarate: You can just do this, no?
```suggestion
								// Flatten the inner blocks, if any.
								$blocks['blocks'] = self::flatten_inner_blocks( $blocks['blocks'], null );
```

src/graphql/graphql-api.php
chriszarate: Nevermind, I was conflating this with querying for `contentNodes`, which is a bit more problematic. I think this is fine.

> this would be what we support - pages and posts that would be made using the content editor.

The `NodeWithContentEditor` interface picks up [any post type that supports the content editor](https://github.com/wp-graphql/wp-graphql/blob/1051357687339e53b7bd19b816c756316dbc8a47/src/Registry/Utils/PostObject.php#L275-L277) regardless of its purpose or visibility. It's worth noting that the default value for `register_post_type#supports` includes `'editor'` (and also many developers and plugin authors don't pay close attention to post type args), so in reality the field will probably get added to a bunch of random plugin-created post types where it might not make sense. But it's fine since the user is unlikely to expose them in GraphQL or query for them. Just FYI.

src/graphql/graphql-api.php
alecgeatches: Huh, weird! We do have `phpcs` configured, and it seems to spot other problems in this file just fine, but it's not enforcing parentheses spacing. Our [phpcs configuration](https://github.com/Automattic/vip-go-mu-plugins/blob/develop/phpcs.xml.dist) is a close relation to [vip-go-mu-plugins's config](https://github.com/Automattic/vip-go-mu-plugins/blob/develop/phpcs.xml.dist) which appears to be roughly the same. Looking into what sniff might be missing here.

README.md
alecgeatches: Agree, addressed in https://github.com/Automattic/vip-block-data-api/pull/51/commits/fac6ebf59e071e5db10234cabe951f7db1d4894f!

README.md
alecgeatches: Fixed in https://github.com/Automattic/vip-block-data-api/commit/fac6ebf59e071e5db10234cabe951f7db1d4894f.

README.md
alecgeatches: Fixed in https://github.com/Automattic/vip-block-data-api/commit/fac6ebf59e071e5db10234cabe951f7db1d4894f.

README.md
alecgeatches: Fixed in https://github.com/Automattic/vip-block-data-api/commit/fac6ebf59e071e5db10234cabe951f7db1d4894f.

README.md
alecgeatches: Good idea! Put a small ["Block Attributes" section](https://github.com/Automattic/vip-block-data-api/blob/fac6ebf59e071e5db10234cabe951f7db1d4894f/README.md#block-attributes) with an example and explaining the trade-off.

README.md
alecgeatches: I think I'm leaning toward keeping these as simplified IDs for the example. Actual IDs returned from GraphQL look super similar (`SUQ6MQ==`, `SUQ6Mg==`, `SUQ6MTY=`) and that gets in the way of demonstrating the connection between blocks and their parents. I tried subbing in some real values, but without easily visible relationships the code is a lot harder to follow:

```json
{
  "data": {
    "post": {
      "blocksData": {
        "blocks": [
          {
            "attributes": [
              {
                "name": "ordered",
                "value": ""
              },
              {
                "name": "values",
                "value": ""
              }
            ],
            "id": "SUQ6MjE=",
            "name": "core/list",
            "innerBlocks": [
              {
                "id": "SUQ6MTk=",
                "name": "core/list-item",
                "parentId": "SUQ6MjE=",
                "attributes": [
                  {
                    "name": "content",
                    "value": "This is item 1 in the list"
                  }
                ]
              },
              {
                "id": "SUQ6MTg=",
                "name": "core/list-item",
                "parentId": "SUQ6MjE=",
                "attributes": [
                  {
                    "name": "content",
                    "value": "This is item 2 in the list"
                  }
                ]
              }
            ]
          },
          {
            "attributes": [
              {
                "name": "value",
                "value": ""
              },
              {
                "name": "citation",
                "value": ""
              }
            ],
            "id": "SUQ6MTY=",
            "name": "core/quote",
            "innerBlocks": [
              {
                "id": "SUQ6MQ==",
                "name": "core/paragraph",
                "parentId": "SUQ6MTY=",
                "attributes": [
                  {
                    "name": "content",
                    "value": "This is a paragraph within a quote"
                  },
                  {
                    "name": "dropCap",
                    "value": ""
                  }
                ]
              }
            ]
          }
        ]
      }
    }
  }
}
```

README.md
alecgeatches: Good point, removed here: https://github.com/Automattic/vip-block-data-api/blob/fac6ebf59e071e5db10234cabe951f7db1d4894f/README.md#graphql-1.

src/graphql/graphql-api.php
alecgeatches: > I guess WPGraphQL is doing the casting to string?

I think you're right! I tested with a block delimiter value like `{"test": 123}`, which is returned as an integer from the parser and REST API. However, once we return it in a `BlockAttribute`, it is converted to a string. Added explicit conversion in https://github.com/Automattic/vip-block-data-api/pull/51/commits/8cdc1e209fb2688dc0930e53ba1e5f0733da18b5 anyway.

README.md
chriszarate: Got it, but the API will return Relay IDs, right? That's important since many client implementations used IDs as cache object identifiers. Using `int`s will result in collisions.

Reasoning makes sense, you could also put the decoded ID in a comment, like

```
"id": "SUQ6MQ==", // "BlockData:1"
```

But good either way

src/graphql/graphql-api.php
alecgeatches: Good catch! I fixed this, tested it manually, and put it in https://github.com/Automattic/vip-block-data-api/pull/51/commits/d578fe470a0ae1248eb0198b682ef76d886bf3ff. I spent some time trying to add a test, but after my attempt I think @ingeniumed has already done a decent job trying to wrap the core logic from the resolver with a test:

https://github.com/Automattic/vip-block-data-api/blob/d578fe470a0ae1248eb0198b682ef76d886bf3ff/src/graphql/graphql-api.php#L220-L233

Our unit testing does not include the WPGraphQL plugin, so the logic that's in this resolver isn't directly tested [in `test_flatten_inner_blocks()`](https://github.com/Automattic/vip-block-data-api/blob/d578fe470a0ae1248eb0198b682ef76d886bf3ff/tests/graphql/test-graphql-api.php#L135). The bug here is in the thin shell of WPGraphQL code around the `flatten_inner_blocks()` function, which we test directly. Ideally we could have some end-to-end tests that include the request through WPGraphQL (like we do [for our REST tests](https://github.com/Automattic/vip-block-data-api/blob/d578fe470a0ae1248eb0198b682ef76d886bf3ff/tests/rest/test-rest-api.php#L130-L138)), but I think the testing here is a good enough start. As long as we catch the interface bugs, like this.

src/graphql/graphql-api.php
alecgeatches: This replacement doesn't work as-is, because `flatten_inner_blocks()` needs the `$parent_id` parameter, which is passed to `innerBlocks` to determine their parent IDs:

https://github.com/Automattic/vip-block-data-api/blob/d578fe470a0ae1248eb0198b682ef76d886bf3ff/src/graphql/graphql-api.php#L116-L119

I think this is due to the design that top level blocks are treated differently as "root" blocks, but their children are flattened beneath them all together. I'm sure it's possible to rewrite `flatten_inner_blocks()` so this is done internally, but I think this works fine.

src/graphql/graphql-api.php
alecgeatches: @ingeniumed Can you explain a bit more about the `stdClass` comment here? I'm not sure where that comes from either.

ingeniumed: So this is because in the contentParser we set the actual attributes to be `(object)[]` which allows it to be given back as empty json in the rest api rather than as an empty array. Due to that, the internal data type comes back as stdClass rather than an array. This workaround is due to that.

I experimented with switching the `(object)[]` to a new ArrayObject but in that case it adds an empty array with key `storage` under the attributes which requires an even worse workaround. That's why it's been done this way.

I tried to search for a better solution but couldn't quite find one and I didn't want to break the rest response or the way the public function works because there are customers already using it.

[This](https://github.com/Automattic/vip-block-data-api/blob/trunk/src/parser/content-parser.php#L268) is the line I'm referencing btw in the contentParser.

ingeniumed: Hopefully that explains it properly

src/graphql/graphql-api.php
alecgeatches: Figured it out. I noticed if I add a missing space a couple of lines up, PHPCS will report it:

```php
$block['id'] = Relay::toGlobalId('ID', wp_unique_id() );
//                               ^ missing space
```

```bash
FILE: src/graphql/graphql-api.php
--------------------------------------------------------------------------
FOUND 1 ERROR AFFECTING 1 LINE
--------------------------------------------------------------------------
 73 | ERROR | [x] Expected 1 spaces after opening parenthesis; 0 found
    |       |     (PEAR.Functions.FunctionCallSignature.SpaceAfterOpenBracket)
```

However, that same `PEAR.Functions.FunctionCallSignature.SpaceAfterOpenBracket` sniff doesn't kick in for `array_map()`. After looking at our configuration, I noticed we have [some sniffs disabled](https://github.com/Automattic/vip-go-mu-plugins/blob/develop/phpcs.xml.dist#L73-L76) related to anonymous functions and function calls. If I remove those, we'll now get multiple errors for the multi-line function call:

```
 72 | ERROR | [x] Opening parenthesis of a multi-line function call must be the last content on the line
    |       |     (PEAR.Functions.FunctionCallSignature.ContentAfterOpenBracket)
 74 | ERROR | [x] Only one argument is allowed per line in a multi-line function call (PEAR.Functions.FunctionCallSignature.MultipleArguments)
 74 | ERROR | [x] Closing parenthesis of a multi-line function call must be on a line by itself (PEAR.Functions.FunctionCallSignature.CloseBracketLine)
```

So, in short, if we want to disable those sniffs to use some standard PHP `array_map( function( ... ) {` practices, we also lose the sniff that tells us about weird spacing after `array_map()`.

Fixed manually.

README.md
alecgeatches: That's correct that these are Relay IDs, and I agree that assuming integers could cause some issues. I tried out your ID commenting suggestion but felt like that also made the example too busy. Settled on a note underneath the example in https://github.com/Automattic/vip-block-data-api/pull/51/commits/90c77dedc6472b01e70226dfd4529b358a011fd4.

src/graphql/graphql-api.php
alecgeatches: @ingeniumed That sounds familiar now! As a side-effect of returning data in JSON via the REST API, if we want to return `attributes: {}` in a response, we need to cast `[]` into an object to avoid returning an array instead. This is just dealing with that unusual data in the GraphQL layer.

alecgeatches:
@chriszarate Thanks a ton for your review! I believe I've addressed all of your suggestions above. Please take another look when you have a chance. Thank you!

src/graphql/graphql-api.php
alecgeatches: Wait a minute, I'm rereading this and I'm not sure why it's not working anymore. Let me take a second look.

src/graphql/graphql-api.php
ingeniumed: Good point, added it in under usage!

src/graphql/graphql-api.php
```diff
@@ -0,0 +1,256 @@
...
+	/**
+	 * Transform the block's format to the format expected by the graphQL API.
+	 *
+	 * @param array $block An associative array of parsed block data with keys 'name' and 'attributes'.
+	 *
+	 * @return array
+	 */
+	public static function transform_block_format( $block ) {
+		// Generate a unique ID for the block.
+		$block['id'] = Relay::toGlobalId( 'ID', wp_unique_id() );
```
chriszarate: The first argument to this function should the type name of the object in question, so `BlockData` in this case. This is what ensures uniqueness across the graph. Also, since `wp_unique_id` just increments integers, you should probably also incorporate the post ID. Otherwise, you risk collisions in client caches which will be really hard to diagnose. (I've seen this happen and it's bewildering when a paragraph from post 2 shows up in post 1).

```suggestion
		$block['id'] = Relay::toGlobalId( 'BlockData', sprintf( "%d:%d", $post->ID, wp_unique_id() ) );
```

alecgeatches: Thank you for clarifying this @chriszarate! Changed in https://github.com/Automattic/vip-block-data-api/pull/51/commits/6e2c3677cb02fcacebe67604a6ff0a78662ef3b1.

chriszarate (commented):
All my feedback has been addressed, but I'll leave it to Bistro to approve :)

ingeniumed:
I'll merge this into our release branch, and we can cut a release later this week via that. Thanks for the amazing feedback @chriszarate, and @smithjw1.

PR #54, opened 2024-01-18
Fix the PHP8+ warnings

ingeniumed:
## Description

This PR will resolve the warnings that show up for PHP 8, as well as upgrade the symfony/css-selector and symfony/dom-crawler packages so that plugin's minimum version is now set to PHP 8.

This also meant marking the postID as a required field, which it already was but this just makes it official.


composer.json
```diff
@@ -15,11 +15,11 @@
...
 	"require": {
 		"masterminds/html5": "^2.8",
-		"symfony/dom-crawler": "^5.4",
-		"symfony/css-selector": "^5.4"
+		"symfony/dom-crawler": "^6.0",
+		"symfony/css-selector": "^6.0"
 	},
 	"require-dev": {
-		"php": ">=7.4",
+		"php": ">=8.0",
```
sanmai:
```suggestion
		"php": ">=8.1",
```
DOM Crawler requires at least 8.1, so it could make sense to bump on our end as well.

vendor/composer/platform_check.php
```diff
@@ -4,8 +4,8 @@
 
 $issues = array();
 
-if (!(PHP_VERSION_ID >= 70205)) {
-    $issues[] = 'Your Composer dependencies require a PHP version ">= 7.2.5". You are running ' . PHP_VERSION . '.';
+if (!(PHP_VERSION_ID >= 80002)) {
+    $issues[] = 'Your Composer dependencies require a PHP version ">= 8.0.2". You are running ' . PHP_VERSION . '.';
```
sanmai, line 8: The same question here; are we sure we can even install it on 8.0.2?

composer.json
```diff
@@ -14,12 +14,12 @@
 		]
 	},
 	"require": {
+		"php": ">=8.0",
 		"masterminds/html5": "^2.8",
 		"symfony/dom-crawler": "^6.0",
 		"symfony/css-selector": "^6.0"
 	},
 	"require-dev": {
-		"php": ">=8.0",
```
sanmai: Do we plan to let others install this as a composer package?

sanmai: Ah. Never mind. I missed the change above.

alecgeatches (approved):
Looks good to me! Going to merge these changes into a new `1.2.0` release.

PR #60, opened 2024-04-08
Add pre- and post- content filter

alecgeatches:
## Description

Add a new filter for modifying post content before it is parsed, and another for after content is parsed:

### `vip_block_data_api__before_parse_post_content`

Modify raw post content before it's parsed by the Block Data API.

```php
/**
 * Filters content before parsing blocks in a post.
 *
 * @param string $post_content The content of the post being parsed.
 * @param int $post_id Post ID associated with the content.
 */
$post_content = apply_filters( 'vip_block_data_api__before_parse_post_content', $post_content, $post_id );
```

---

### `vip_block_data_api__after_parse_blocks`

Modify the Block Data API REST endpoint response.

```php
/**
 * Filters the API result before returning parsed blocks in a post.
 *
 * @param string $result The successful API result, contains 'blocks' key with an array
 *                       of block data, and optionally 'warnings' and 'debug' keys.
 * @param int $post_id Post ID associated with the content.
 */
$result = apply_filters( 'vip_block_data_api__after_parse_blocks', $result, $post_id );
```

## Steps to Test

Tests have been added both of the new filters in `tests/parser/test-parser-filters.php`. To test:

```bash
$ wp-env start
$ composer install
$ composer run test

> wp-env run tests-cli --env-cwd=wp-content/plugins/vip-block-data-api ./vendor/bin/phpunit
ℹ Starting './vendor/bin/phpunit' on the tests-cli container. 

....................................................              52 / 52 (100%)

Time: 00:00.270, Memory: 56.50 MB

OK (52 tests, 122 assertions)
✔ Ran `./vendor/bin/phpunit` in 'tests-cli'. (in 3s 227ms)
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -1218,6 +1218,64 @@ Direct block HTML can be accessed through `$block['innerHTML']`. This may be use
 
 For another example of how this filter can be used to extend block data, we have implemented a default image block filter in [`src/parser/block-additions/core-image.php`][repo-core-image-block-addition]. This filter is automatically called on `core/image` blocks to add `width` and `height` to image attributes.
 
+---
+
+### `vip_block_data_api__before_parse_post_content`
+
+Modify raw post content before it's parsed by the Block Data API.
+
+```php
+/**
+ * Filters content before parsing blocks in a post.
+ *
+ * @param string $post_content The content of the post being parsed.
+ * @param int $post_id Post ID associated with the content.
+ */
+$post_content = apply_filters( 'vip_block_data_api__before_parse_post_content', $post_content, $post_id );
+```
+
+---
+
+### `vip_block_data_api__after_parse_blocks`
+
+Modify the Block Data API REST endpoint response.
+
+```php
+/**
+ * Filters the API result before returning parsed blocks in a post.
+ *
+ * @param string $result The successful API result, contains 'blocks' key with an array
+ *                       of block data, and optionally 'warnings' and 'debug' keys.
+ * @param int $post_id Post ID associated with the content.
+ */
+$result = apply_filters( 'vip_block_data_api__after_parse_blocks', $result, $post_id );
+```
+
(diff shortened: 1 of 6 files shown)


alecgeatches:
@smithjw1 Could I please have a review from you on this, especially on the README documentation for the new filters? Thank you!

README.md
```diff
@@ -1218,6 +1218,64 @@ Direct block HTML can be accessed through `$block['innerHTML']`. This may be use
 
 For another example of how this filter can be used to extend block data, we have implemented a default image block filter in [`src/parser/block-additions/core-image.php`][repo-core-image-block-addition]. This filter is automatically called on `core/image` blocks to add `width` and `height` to image attributes.
 
+---
+
+### `vip_block_data_api__before_parse_post_content`
+
+Modify raw post content before it's parsed by the Block Data API.
```
smithjw1: Like you do with the other filter, I might mention that this is the very first thing called so that you are getting exactly what is in the WP database.

I also like the example in the other one, so we could add one here. The example helps people understand why they might use it.

We should also add a warning that manipulating the content directly can cause block parsing errors, and they should be careful with this filter. Here's the language I would use:

> Block Markup is sensitive to changes, even changes in whitespace. We've added this filter to make the plugin flexible, but any changes to the `post_content` should be done with extreme care. Strongly consider adding tests to any filter.

smithjw1 (changes requested):
I've made a few comments on the Readme. I think we should have someone else review the code, it looks good to me, but I didn't dig too much into the tests.

README.md
alecgeatches: Great notes, especially the warning. Will add!

alecgeatches: @smithjw1 Added an before_parse example along with a warning as you described above. Thank you!

smithjw1 (dismissed):
Wonderfully done @alecgeatches

alecgeatches:
I'm going to merge this as the code changes are low-risk filter additions and also include tests.

PR #61, opened 2024-04-24
Fix "Array" values in GraphQL block array data

### LINKED ISSUE #53: WPGraphQL attributes missing with core/table array attributes
WPGraphQL results for table attributes with array values have `Array` strings and are missing children.

## Reproduction

1. Create a post containing a table, e.g.

    image
    
    ```html
    <!-- wp:table -->
    <figure class="wp-block-table"><table><tbody><tr><td>Header A</td><td>Header B</td></tr><tr><td>Value 1</td><td>Value 2</td></tr></tbody></table></figure>
    <!-- /wp:table -->
    ```

2. Using the created post's ID, query for `blocks`/`innerBlocks` data:

    ```graphql
    query NewQuery {
      post(id: 431, idType: DATABASE_ID) {
        blocksData {
          blocks {
            attributes {
              name
              value
            }
            id
            name
            innerBlocks {
              attributes {
                name
                value
              }
              id
              name
              parentId
            }
          }
        }
      }
    }
    ```

3. See "Array" in results:

    ```js
    {
      "data": {
        "post": {
          "blocksData": {
            "blocks": [
              {
                "attributes": [
                  /* ... */
                  {
                    "name": "head",
                    "value": "Array"
                  },
                  {
                    "name": "body",
                    "value": "Array"
                  },
                  {
                    "name": "foot",
                    "value": "Array"
                  }
                ],
                "name": "core/table",
                "innerBlocks": null
              }
            ]
          }
        }
      }
    }
    ```

    Note that the `head`, `body`, and `foot` attributes all have a value of `"Array"`, rather than an actual array. It's not possible to reconstruct the table.

## Expected results

The REST API returns array values as expected:

```js
{
  "blocks": [
    {
      "name": "core/table",
      "attributes": {
        /* ... */
        "head": [],
        "body": [
          {
            "cells": [
              {
                "content": "Header A",
                "tag": "td"
              },
              {
                "content": "Header B",
                "tag": "td"
              }
            ]
          },
          {
            "cells": [
              {
                "content": "Value 1",
                "tag": "td"
              },
              {
                "content": "Value 2",
                "tag": "td"
              }
            ]
          }
        ],
        "foot": []
      }
    }
  ]
}
```

## Fixing it

At the time of writing, WPGraphQL `BlockAttribute` is defined as a `{ name: String, value: String }` combination:

```php
register_graphql_object_type(
    'BlockAttribute',
    [
        'description' => 'Block attribute',
        'fields'      => [
            'name'  => [
                'type'        => [ 'non_null' => 'String' ],
                'description' => 'Block data attribute name',
            ],
            'value' => [
                'type'        => [ 'non_null' => 'String' ],
                'description' => 'Block data attribute value',
            ],
        ],
    ],
);
```

Possible fixes could be to make the value support multiple types (`String|Array`) if that's possible in GraphQL, or we could continue to use the `String` type and put complex data types as JSON, e.g.

```js
"blocksData": {
  "blocks": [
    {
      "attributes": [
        /* ... */
        {
          "name": "head",
          "value": "[]"
        },
        {
          "name": "body",
          "value": "{\"cells\": [{\"content\": \"Header A\", \"tag\": \"td\"}, {\"content\": \"Header B\",\"tag\": \"td\"}]}"
        },
        {
          "name": "foot",
          "value": "[]"
        }
      ],
      "name": "core/table",
      "innerBlocks": null
    }
  ]
}
```

alecgeatches:
## Description

See #53 for a description of the problem. In short, the code that maps block attributes into GraphQL types coerces the `value` property into a string:

https://github.com/Automattic/vip-block-data-api/blob/f89ed3c742e8d56dd4cda55056dc738422ffa33e/src/graphql/graphql-api.php#L93-L102

This causes an `'Array'` string to replace the actual value, which makes `array` data inaccessible through WPGraphQL. Additionally, this coercion produces a notice:

```
PHP Warning:  Array to string conversion in /.../vip-block-data-api/src/graphql/graphql-api.php on line 97
```

## Why JSON?

The fix in this PR checks for arrays in the `value` property, and converts those to a JSON string representation. This was selected instead of per-block custom properties or union types.

`array`-source values can represent arbitrarily complex data which could require adding a new type for each block. For example, the `head`, `body`, and `foot` of a table follow a structured `cell`-based pattern:

```json
{
    "name": "core/table",
    "attributes": {
        "hasFixedLayout": false,
        "head": [
            { "cells": [
                { "content": "Header A", "tag": "th" },
                { "content": "Header B", "tag": "th" }
            ] }
        ],
        "body": [
            { "cells": [
                { "content": "Value A", "tag": "td" },
                { "content": "Value B", "tag": "td" }
            ] },
            { "cells": [
                { "content": "Value C", "tag": "td" },
                { "content": "Value D", "tag": "td" }
            ] }
        ],
        "foot": [
            { "cells": [
                { "content": "Footer A", "tag": "td"},
                { "content": "Footer B", "tag": "td" }
            ] }
        ]
    }
}
```

That's because `core/table` uses a [`query` with a known structure](https://github.com/WordPress/gutenberg/blob/74a06c7/packages/block-library/src/table/block.json#L24) to fill those values.

However, other uses of `array` values, like the [`tracks` attribute of `core/video`](https://github.com/WordPress/gutenberg/blob/trunk/packages/block-library/src/video/block.json#L72-L79) use a freeform structure:

```json
{
    "name": "core/video",
    "attributes": {
        "src": "http://my.site/wp-content/uploads/my-video.mp4",
        "tracks": [
            {
                "src": "http://my.site/wp-content/uploads/track-en.vtt",
                "label": "track-en",
                "srcLang": "en"
            },
            {
                "src": "http://my.site/wp-content/uploads/track-fr.vtt",
                "label": "track-fr",
                "srcLang": "fr",
                "kind": "subtitles"
            }
        ]
    }
}
```

If we tried to cover known core `array` types with custom types, this still wouldn't solve the problem for custom blocks. JSON representation will allow the GraphQL backend to model arbitrarily complex nested types without needing to know their structure beforehand.

The primary downside is that consumers of the data will be required to know that the block contains JSON-encoded information for `array`-type values.

However, that doesn't change the existing model for querying. All attribute values are already treated as `String` type [as a trade-off for convenience](https://github.com/Automattic/vip-block-data-api/blob/b0fc4b7/README.md?plain=1#L377). Also, given that array values were completely clobbered into an `"Array"` string previously, this is an improvement.

## Steps to Test

A test using `core/table` was added in [`test_array_data_in_attribute()`](https://github.com/Automattic/vip-block-data-api/blob/7bfc3c684112c413ee2cbaa2192722bcb0aa9675/tests/graphql/test-graphql-api.php#L272). To run tests:

1. Check out PR.
1. Run `wp-env start`
1. Run `composer install`
1. Run `composer run test`

Fixes #53

### DIFF
--- a/src/graphql/graphql-api.php
+++ b/src/graphql/graphql-api.php
@@ -91,12 +91,7 @@ class GraphQLApi {
 			unset( $block['attributes'] );
 		} elseif ( isset( $block['attributes'] ) && ! empty( $block['attributes'] ) ) {
 			$block['attributes'] = array_map(
-				function ( $name, $value ) {
-					return [
-						'name'  => $name,
-						'value' => strval( $value ),
-					];
-				},
+				[ __CLASS__, 'get_block_attribute_pair' ],
 				array_keys( $block['attributes'] ),
 				array_values( $block['attributes'] )
 			);
@@ -251,6 +246,23 @@ class GraphQLApi {
 				'resolve'     => [ __CLASS__, 'get_blocks_data' ],
 			]
 		);
+	}
+
+	/**
+	 * Given a block attribute name and value, return a BlockAttribute array.
+	 *
+	 * @return array
+	 */
+	public static function get_block_attribute_pair( $name, $value ) {
+		// Unknown array types (table cells, for example) are encoded as JSON strings.
+		if ( is_array( $value ) ) {
+			$value = wp_json_encode( $value );
+		}
+
+		return [
+			'name'  => $name,
+			'value' => strval( $value ),
+		];
 	}
 }
 
--- a/src/graphql/graphql-api.php
+++ b/src/graphql/graphql-api.php
@@ -25,7 +25,7 @@ class GraphQLApi {
 	}
 
 	/**
-	 * Extract the blocks data for a post, and return back in the format expected by the graphQL API.
+	 * Extract the blocks data for a post, and return back in the format expected by the GraphQL API.
 	 *
 	 * @param  \WPGraphQL\Model\Post $post_model Post model for post.
 	 *
@@ -154,13 +154,17 @@ class GraphQLApi {
 			[
 				'description' => 'Block attribute',
 				'fields'      => [
-					'name'  => [
+					'name'               => [
 						'type'        => [ 'non_null' => 'String' ],
 						'description' => 'Block data attribute name',
 					],
-					'value' => [
+					'value'              => [
 						'type'        => [ 'non_null' => 'String' ],
 						'description' => 'Block data attribute value',
+					],
+					'isValueJsonEncoded' => [
+						'type'        => [ 'non_null' => 'Boolean' ],
+						'description' => 'True if value is a complex JSON-encoded field. This is used to encode attribute types like arrays and objects.',
 					],
 				],
 			],
@@ -258,13 +262,17 @@ class GraphQLApi {
 	 */
 	public static function get_block_attribute_pair( $name, $value ) {
 		// Unknown array types (table cells, for example) are encoded as JSON strings.
-		if ( is_array( $value ) ) {
-			$value = wp_json_encode( $value );
+		$is_value_json_encoded = false;
+
+		if ( ! is_scalar( $value ) ) {
+			$value                 = wp_json_encode( $value );
+			$is_value_json_encoded = true;
 		}
 
 		return [
-			'name'  => $name,
-			'value' => strval( $value ),
+			'name'               => $name,
+			'value'              => strval( $value ),
+			'isValueJsonEncoded' => $is_value_json_encoded,
 		];
 	}
 }
--- a/src/graphql/graphql-api.php
+++ b/src/graphql/graphql-api.php
@@ -251,6 +251,9 @@ class GraphQLApi {
 	/**
 	 * Given a block attribute name and value, return a BlockAttribute array.
 	 *
+	 * @param string $name The name of the block attribute.
+	 * @param mixed  $value The value of the block attribute.
+	 *
 	 * @return array
 	 */
 	public static function get_block_attribute_pair( $name, $value ) {
--- a/README.md
+++ b/README.md
@@ -376,6 +376,85 @@ The attributes of a block in GraphQL are available in a list of `name` / `value`
 
 This is used instead of a key-value structure. This is a trade-off that makes it easy to retrieve block attributes without specifying the the block type ahead of time, but attribute type information is lost.
 
+
+#### Complex attributes
+
+Some block attributes contain arrays or complex nested values. Demonstrated below, [the `core/table` block uses an array of objects][gutenberg-code-table-body] to represent head, body, and footer cell contents. The GraphQL Block Data API implementation represents these attributes as JSON-encoded strings along with the `isValueJsonEncoded` boolean field. When `isValueJsonEncoded` is `true`, an attribute's value must be JSON decoded to get the original complex value.
+
+For example, using this table:
+
+![Example core/table block with a two header cells and two body cells][media-example-table]
+
+We can query for attributes along with the `isValueJsonEncoded` field in a GraphQL query:
+
+```graphql
+query PostQuery {
+  post(id: 1, idType: DATABASE_ID) {
+    blocksData {
+      blocks {
+        attributes {
+          name
+          value
+          isValueJsonEncoded
+        }
+        id
+        name
+        innerBlocks {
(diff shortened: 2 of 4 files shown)


src/graphql/graphql-api.php
```diff
@@ -252,6 +247,26 @@ public static function register_types() {
...
+	 * Given a block attribute name and value, return a BlockAttribute array.
+	 *
+	 * @param string $name The name of the block attribute.
+	 * @param mixed  $value The value of the block attribute.
+	 *
+	 * @return array
+	 */
+	public static function get_block_attribute_pair( $name, $value ) {
+		// Unknown array types (table cells, for example) are encoded as JSON strings.
+		if ( is_array( $value ) ) {
```
Zamfi99: What do you think about using `!is_scalar( $value )` instead?

src/graphql/graphql-api.php
```diff
@@ -252,6 +247,26 @@ public static function register_types() {
...
+			$value = wp_json_encode( $value );
+		}
+
+		return [
```
Zamfi99, line 272: I'm thinking if it would make sense to specify the actual type of `$value` (which would be the best approach imo), or at least specifying that the content is encoded (we may run into performance issues by trying to decode every attribute in frontend)

alecgeatches:
Thanks a ton for your input, @Zamfi99! We [just pushed a change](https://github.com/Automattic/vip-block-data-api/pull/61/commits/9ac997108b1fb6fbc22e1caedf65b906d97d1140) that adds a `isValueJsonEncoded` boolean to the `attributes` query. Here's an example with a table:

image

*Query*

```graphql
query MyPostQuery {
  post(id: $ID, idType: DATABASE_ID) {
    blocksData {
      blocks {
        attributes {
          name
          value
          isValueJsonEncoded      " <-- New attribute "
        }
        id
        name
        innerBlocks {
          attributes {
            name
            value
            isValueJsonEncode     " <-- New attribute "
          }
          id
          name
          parentId
        }
      }
    }
  }
}
```

→ *Result*

```json
{
  "data": {
    "post": {
      "blocksData": {
        "blocks": [
          {
            "attributes": [
              {
                "name": "hasFixedLayout",
                "value": "",
                "isValueJsonEncoded": false
              },
              {
                "name": "head",
                "value": "[{\"cells\":[{\"content\":\"Header A\",\"tag\":\"th\"},{\"content\":\"Header B\",\"tag\":\"th\"}]}]",
                "isValueJsonEncoded": true
              },
              {
                "name": "body",
                "value": "[{\"cells\":[{\"content\":\"Value A\",\"tag\":\"td\"},{\"content\":\"Value B\",\"tag\":\"td\"}]},{\"cells\":[{\"content\":\"Value C\",\"tag\":\"td\"},{\"content\":\"Value D\",\"tag\":\"td\"}]}]",
                "isValueJsonEncoded": true
              },
              {
                "name": "foot",
                "value": "[{\"cells\":[{\"content\":\"Footer A\",\"tag\":\"td\"},{\"content\":\"Footer B\",\"tag\":\"td\"}]}]",
                "isValueJsonEncoded": true
              }
            ],
            "id": "QmxvY2tEYXRhOjQ1NTox",
            "name": "core/table",
            "innerBlocks": null
          }
        ]
      }
    }
  }
}
```

We liked the idea of an opt-in `isValueJsonEncoded` attribute. This solves a couple of problems:

1. You can avoid decoding every attribute by querying for `isValueJsonEncoded` along with attribute `name` and `value`, and only do extra decoding work when this value is `true`.
2. Folks who don't use blocks with array values can continue to query for simple `name`/`value` pairs and can use this feature optionally.

We also chose not to add the attribute's type as a queryable option, because we're actively trying to keep the GraphQL querying options simple. I also think it feels confusing to not use a "real" GraphQL type and specify a type another attribute instead. This is the same trade-off we made for [original attribute typing as strings](https://github.com/Automattic/vip-block-data-api/blob/b0fc4b7/README.md?plain=1#L377). I think the `isValueJsonEncoded` attribute is a good middle ground for attributes that can't be easily represented as a simple string.

Let me know what you think! We're look to get this shipped out this week. Thank you for your input!

alecgeatches:
@ingeniumed Would appreciate your review here as well, thank you!

ingeniumed (dismissed):

ingeniumed:
Forgot to mention that the plugin version should be bumped

ingeniumed (approved):

PR #63, opened 2024-05-26
Run the unit tests for the lowest and latest versions of wp/php

ingeniumed:
## Description

This will update the PHP unit tests workflow to run for the lowest WP/PHP version we support and the latest combo as well. That way at all times we are ensuring our code is working as expected. It also updates some of the actions that are used, to take advantage of the caching/etc improvements.


alecgeatches:
@ingeniumed The tests have been fixed, would love a review! The main issue here is that some tests were using core blocks directly, and as their attributes changed between WordPress updates it caused direct attribute comparisons to fail. A lot of tests were immune to this problem by registering their own block definitions for tests, like attribute tests:

https://github.com/Automattic/vip-block-data-api/blob/32cda50b4c545ac8b70f62669f31a96f43a4d963/tests/parser/sources/test-source-html.php#L16-L22

However, REST API and GraphQL tests were using core blocks directly:

https://github.com/Automattic/vip-block-data-api/blob/fc73fc5b85c37901e6270ee92a88f1f847307d28/tests/rest/test-rest-api.php#L39-L51

...

https://github.com/Automattic/vip-block-data-api/blob/fc73fc5b85c37901e6270ee92a88f1f847307d28/tests/rest/test-rest-api.php#L74-L81

The problem is that when a `block.json` attribute is added or changed in a new WordPress version, any test using those blocks and attribute comparison would fail. This was a bit more difficult to fix than the first example test because REST API and GraphQL tests that fire off a request and don't have an easy way to specify a custom registry.

In the changes, I've added a couple of `register_global_block_with_attributes()` methods (in REST and GraphQL tests respectively) that modify the "real" registry, and then automatically unregister custom blocks after the test completes. This allows us to use mocked core blocks without having a direct dependency on the block's definition in WordPress.

This should fix all test breaks related to core block attribute changes, and make future WordPress testing for this plugin less brittle.

vip-block-data-api.php
```diff
@@ -5,8 +5,8 @@
  * Description: Access Gutenberg block data in JSON via the REST API.
  * Author: WordPress VIP
  * Text Domain: vip-block-data-api
- * Version: 1.2.3
- * Requires at least: 5.9.0
+ * Version: 1.2.4
+ * Requires at least: 6.0
  * Tested up to: 6.4
```
ingeniumed, line 11: Should update this to 6.6 btw

ingeniumed (commented):
Great improvements to the tests! Just a minor item to tweak before this can be 🚢

PR #71, opened 2024-07-30
Fix asterisk attribute selector

alecgeatches:
## Description

See https://github.com/Automattic/vip-block-data-api/pull/65. When attributes use a `*` selector, this can refer to the wrapping `<body>` element used in the parser and can fail to source attributes correctly.

This PR changes the parser to automatically start with the `children()` of `body` so that the wrapping HTML context is no longer relevant. Note that this also requires the `raw` source to use `outerHtml()` instead of `html()` as the body tag is no longer part of the context. I've tested the `raw` change does not break manual testing with the `core/html` block (at root level and nested), as well as the automated tests already in place.

One new test has been added for the `*` anchor selector case.

## Steps to Test

1. Check out PR.
1. Run `wp-env start`
1. Run `composer install`
1. Run `composer run test`

### DIFF
--- a/src/parser/content-parser.php
+++ b/src/parser/content-parser.php
@@ -240,7 +240,7 @@ class ContentParser {
 			$crawler = new Crawler( sprintf( '<!doctype html><html><body>%s</body></html>', $block['innerHTML'] ) );
 
 			// Enter the <body> tag for block parsing.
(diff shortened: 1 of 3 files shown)


Zamfi99:
lgtm, thanks @alecgeatches!

PR #74, opened 2024-08-23
Add block binding support

chriszarate:
## Description

This PR adds support for [block bindings](https://make.wordpress.org/core/2024/03/06/new-feature-the-block-bindings-api/), introduced in WordPress 6.5.

The changes here are relatively small, although they are accompanied by some test improvements and cleanup, so the diff is a bit larger than it otherwise would be. Breakdown of commits:

### New functionality

- [Update parser to leverage core block rendering](https://github.com/Automattic/vip-block-data-api/commit/1b6984208eeab86414141c14bf8555d6b70ddd9e): This is the most impactful commit:
  - In order to support block binding without copying many lines of code from core, we create a block instance and render it (including all of its inner blocks), then access the computed attributes and inner HTML. This lets Core do the heavy lifting.
  - Then, we walk the tree and apply our logic to extract sourced attributes.

### Test improvements

No test data was changed in this PR (only ordering).

- [Simplify and improve RegistryTestCase](https://github.com/Automattic/vip-block-data-api/commit/40ff45772a19f85007ab8138265b36829cdd09f2): Removes the unnecessary distinction between "global" and normal block type registration and delegates more work to Core classes.
- [Update tests to match alphabetical key order](https://github.com/Automattic/vip-block-data-api/commit/ccbc687f353b762d9cd91c16b9d7e3e239a63889). Sorting the array keys of attributes more less brittle tests. 

### Additional test cases

- [Add failing test for single block binding](https://github.com/Automattic/vip-block-data-api/commit/b974c6dd807b97ee1d3635bd1041297114d36022).
- [Add (now passing) test for block binding with context](https://github.com/Automattic/vip-block-data-api/commit/f880209e70fa47c6275fe5ef4ac57eb3f50fdbb7)
No test data was changed, only order.
- [Add additional test cases](https://github.com/Automattic/vip-block-data-api/commit/12f50d9f724b6c3e276fcbf1dae0d3ef145bba06)
- [Add test for block binding with post context](https://github.com/Automattic/vip-block-data-api/commit/510ca2ef9f12c339e33d1f4c6cb18cacc317247c)
- [Skip block binding tests when WordPress < 6.5](https://github.com/Automattic/vip-block-data-api/pull/74/commits/bfec40a68c12f4922c501d7ed5948a7ca8af3f2d)

### DIFF
--- a/src/parser/content-parser.php
+++ b/src/parser/content-parser.php
@@ -61,21 +61,28 @@ class ContentParser {
 	 * Filter out a block from the blocks output based on:
 	 *
 	 * - include parameter, if it is set or
-	 * - exclude parameter, if it is set.
+	 * - exclude parameter, if it is set or
+	 * - whether it is an empty whitespace block
 	 *
 	 * and finally, based on a filter vip_block_data_api__allow_block
 	 *
-	 * @param array  $block Current block.
-	 * @param string $block_name Name of the block.
-	 * @param array  $filter_options Options to be used for filtering, if any.
+	 * @param WP_Block $block Current block.
+	 * @param array    $filter_options Options to be used for filtering, if any.
 	 *
 	 * @return bool true, if the block should be included or false otherwise
 	 *
 	 * @access private
 	 */
-	protected function should_block_be_included( $block, $block_name, $filter_options ) {
+	protected function should_block_be_included( WP_Block $block, array $filter_options ) {
+		$block_name        = $block->name;
 		$is_block_included = true;
 
+		// Whitespace blocks are always excluded.
+		$is_whitespace_block = null === $block_name && empty( trim( $block->inner_html ) );
+		if ( $is_whitespace_block ) {
+			return false;
+		}
+		
 		if ( ! empty( $filter_options['include'] ) ) {
 			$is_block_included = in_array( $block_name, $filter_options['include'] );
 		} elseif ( ! empty( $filter_options['exclude'] ) ) {
@@ -86,11 +93,11 @@ class ContentParser {
 		 * Filter out blocks from the blocks output
 		 *
 		 * @param bool   $is_block_included True if the block should be included, or false to filter it out.
-		 * @param string $block_name   Name of the parsed block, e.g. 'core/paragraph'.
-		 * @param string $block         Result of parse_blocks() for this block.
+		 * @param string $block_name    Name of the parsed block, e.g. 'core/paragraph'.
+		 * @param array  $block         Result of parse_blocks() for this block.
 		 *                              Contains 'blockName', 'attrs', 'innerHTML', and 'innerBlocks' keys.
 		 */
-		return apply_filters( 'vip_block_data_api__allow_block', $is_block_included, $block_name, $block );
+		return apply_filters( 'vip_block_data_api__allow_block', $is_block_included, $block_name, $block->parsed_block );
 	}
 
 	/**
@@ -137,10 +144,6 @@ class ContentParser {
 			$post_content = apply_filters( 'vip_block_data_api__before_parse_post_content', $post_content, $post_id );
 
 			$blocks = parse_blocks( $post_content );
-			$blocks = array_values( array_filter( $blocks, function ( $block ) {
-				$is_whitespace_block = ( null === $block['blockName'] && empty( trim( $block['innerHTML'] ) ) );
-				return ! $is_whitespace_block;
-			} ) );
 
 			$sourced_blocks = array_map( function ( $block ) use ( $filter_options ) {
 				// Render the block, then walk the tree using source_block to apply our
@@ -228,7 +231,7 @@ class ContentParser {
 	protected function source_block( WP_Block $block, array $filter_options ) {
 		$block_name = $block->name;
 
-		if ( ! $this->should_block_be_included( $block->parsed_block, $block_name, $filter_options ) ) {
+		if ( ! $this->should_block_be_included( $block, $filter_options ) ) {
 			return null;
 		}
 
--- a/src/parser/content-parser.php
+++ b/src/parser/content-parser.php
@@ -252,9 +252,9 @@ class ContentParser {
 		$inner_blocks = iterator_to_array( $block->inner_blocks );
 
 		// Recursively iterate over inner blocks.
-		$sourced_inner_blocks = array_filter( array_map( function ( $inner_block ) use ( $filter_options ) {
+		$sourced_inner_blocks = array_values( array_filter( array_map( function ( $inner_block ) use ( $filter_options ) {
 			return $this->source_block( $inner_block, $filter_options );
-		}, $inner_blocks ) );
+		}, $inner_blocks ) ) );
 
 		// This empty check is not strictly necessary, but previous versions did
 		// not set innerBlocks if the array was empty.
--- a/src/parser/content-parser.php
+++ b/src/parser/content-parser.php
@@ -256,8 +256,7 @@ class ContentParser {
 			return $this->source_block( $inner_block, $filter_options );
 		}, $inner_blocks ) ) );
 
-		// This empty check is not strictly necessary, but previous versions did
-		// not set innerBlocks if the array was empty.
+		// Only set innerBlocks if entries are present to match prior version behavior.
 		if ( ! empty( $sourced_inner_blocks ) ) {
 			$sourced_block['innerBlocks'] = $sourced_inner_blocks;
 		}
--- a/src/parser/content-parser.php
+++ b/src/parser/content-parser.php
@@ -11,7 +11,7 @@ defined( 'ABSPATH' ) || die();
 
 use Throwable;
 use WP_Error;
-use WP_Block_Type;
+use WP_Block;
 use WP_Block_Type_Registry;
 use Symfony\Component\DomCrawler\Crawler;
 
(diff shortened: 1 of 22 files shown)


tests/parser/test-block-bindings.php
```diff
@@ -0,0 +1,280 @@
...
+				],
+			],
+		];
+
+		$content_parser = new ContentParser( $this->get_block_registry() );
+		$blocks         = $content_parser->parse( $html );
+
+		$this->assertArrayHasKey( 'blocks', $blocks, sprintf( 'Unexpected parser output: %s', wp_json_encode( $blocks ) ) );
+
+		// Because this test uses a core block, we use assertArraySubset to avoid brittle
```
alecgeatches: Could you expand this comment to include why you are using a core block? My understanding is that `core/paragraph` is special and custom blocks do not support content bindings, but it could also just be because you wanted to test a core block.

tests/registry-test-case.php
```diff
@@ -12,50 +13,48 @@
...
 	use ArraySubsetAsserts;
 
-	protected $registry;
-	protected $globally_registered_blocks = [];
-
-	protected function setUp(): void {
-		parent::setUp();
+	protected function tearDown(): void {
+		// Unregister non-core blocks.
+		$block_registry = WP_Block_Type_Registry::get_instance();
```
alecgeatches, line 18: This is a clever way to persist block registrations during GraphQL tests, and also support core blocks for testing. Nice! I wasn't a big fan of the `global` set of registration functions, and this nicely hides those details away.

src/parser/content-parser.php
```diff
@@ -61,21 +63,28 @@ public function __construct( $block_registry = null ) {
...
 	 *
 	 * @access private
 	 */
-	public function should_block_be_included( $block, $block_name, $filter_options ) {
+	protected function should_block_be_included( WP_Block $block, array $filter_options ) {
+		$block_name        = $block->name;
 		$is_block_included = true;
 
+		// Whitespace blocks are always excluded.
+		$is_whitespace_block = null === $block_name && empty( trim( $block->inner_html ) );
```
alecgeatches, line 83: It looks like this logic was moved from `parse()`. Could you add a test ensuring whitespace blocks are removed from the output, maybe in `test-content-parser.php`?

src/parser/content-parser.php
```diff
@@ -189,37 +196,114 @@ public function parse( $post_content, $post_id = null, $filter_options = [] ) {
...
+		// WP_Block#inner_blocks can be an array or WP_Block_List (iterable).
+		$inner_blocks = iterator_to_array( $block->inner_blocks );
+
+		// Recursively iterate over inner blocks.
+		$sourced_inner_blocks = array_values( array_filter( array_map( function ( $inner_block ) use ( $filter_options ) {
+			return $this->source_block( $inner_block, $filter_options );
+		}, $inner_blocks ) ) );
+
+		// This empty check is not strictly necessary, but previous versions did
+		// not set innerBlocks if the array was empty.
```
alecgeatches: Nit, I'd remove the "This empty check is not strictly necessary" part:

```suggestion
		// Only set innerBlocks if entries are present to match prior version behavior
```

alecgeatches (dismissed):
I think this looks and works really well! I especially like the improvements to tests. 

There's quite a bit that's changed in the core block rendering implementation, but with our existing test suite and the new ones you've added I'm confident that we're not breaking anything here.

I added some minor comments about comments, but the functionality looks great to me. I'd also like @ingeniumed to have a look soon as well.

tests/parser/test-block-bindings.php
chriszarate: It's the former, block bindings are not yet supported outside of a few core blocks. Clarified in https://github.com/Automattic/vip-block-data-api/pull/74/commits/a67d7e5efd8aa402c97b7d56cf0e15fc1f32433f

src/parser/content-parser.php
chriszarate: In reality just about every unit test is testing for this since we are indenting the blocks in our HTML fixtures, which results in whitespace blocks. But I added an explicit test in https://github.com/Automattic/vip-block-data-api/pull/74/commits/9eb3e2cd9f80a72e9b2d8230a6b1918658d2e710

src/parser/content-parser.php
chriszarate: I took your suggestion in https://github.com/Automattic/vip-block-data-api/pull/74/commits/674d006f58d1a1ae8cb870f31c46035ce26c5752. 

I tend to think conditional API properties are an anti-pattern. As an API consumer I'd much rather see a consistent response shape. 

I doubt this change would be a breaking one for any users, but it would require updating many test fixtures, so that's why I avoided doing so in this PR. But I'd be happy to do so if it is viewed as an improvement.

alecgeatches: > I tend to think conditional API properties are an anti-pattern. As an API consumer I'd much rather see a consistent response shape.

That's a good point I hadn't considered. For now, let's keep the response shape as-is in case folks have strict equality checks in consumers like `if ( block.innerBlocks === undefined )`. I'd like to keep this change in mind if/when we make a breaking change to a `v2` API.

src/parser/content-parser.php
alecgeatches: Thank you!

alecgeatches:
@chriszarate The discussion in https://github.com/Automattic/vip-block-data-api/pull/74#discussion_r1731992594 made me think about a potential version bump to `v2`, as in `vip-block-data-api/v2/...`. We recently did a similar thing for the GraphQL endpoint with `blockDataV2`: https://github.com/Automattic/vip-block-data-api/pull/72.

Are there any changes in this PR or #75 that would justify a version upgrade? It seems like we're mostly adding more data to the response with the original binding data + actual computed binding results, which isn't necessarily a breaking change but could be. What do you think? This would entail some more work to duplicate some tests and separate a second parser class, but it may be an easier path and allow some other minor upgrades [like we discussed](https://github.com/Automattic/vip-block-data-api/pull/74#discussion_r1731992594).

chriszarate:
> Are there any changes in this PR or https://github.com/Automattic/vip-block-data-api/pull/75 that would justify a version upgrade?

There aren't any breaking changes to the response shape—no new properties or changes to existing ones. The functional changes (supporting block bindings and synced patterns) reflect the current approach and behavior of the plugin, and just enhance the existing response.

Anything can be a surprise of course, but let's use some hypotheticals to guide us. All of these assume we release these two PRs and _don't_ bump to `v2`:

### Block bindings

The effect of block bindings in this plugin is potentially updated string attribute values for specific core blocks when block bindings are in use. There is currently no UI for block bindings in the block editor and very little adoption, so anyone using them must write and deploy code to provide them. If a Block Data API user _also_ uses block bindings, they are certainly aware that they are not supported—and probably frustrated.

1. I am a Block Data API consumer and I don't use block bindings. Upgrading has no effect.
2. I am a Block Data API consumer and I use block bindings. I created my own implementation to resolve the bindings using the returned `metadata` attributes. Upgrading has no effect since my implementation governs and overrides. I can remove my implementation at any time of my choosing.
3. I am a Block Data API consumer and I use block bindings. The returned data is incorrect, but I have decided to use it anyway. Upgrading has an effect by providing the correct data, and I consider this a breaking change.

Scenarios 1 and 2 are not breaking changes. A breaking change via scenario 3 is _possible_, but I struggle to imagine a realistic example.

### Synced patterns

As shown in the [PR description](https://github.com/Automattic/vip-block-data-api/pull/75), the effect of synced pattern support is a populated `innerBlocks` property for `core/block` blocks.

Synced patterns do have UI in the block editor, but Block Data API users are almost certainly aware that they are not supported.

1. I am a Block Data API consumer and I don't use synced patterns. Upgrading has no effect.
2. I am a Block Data API consumer and I use synced patterns. I created my own implementation to resolve the synced patterns using the returned `ref` attribute. Upgrading has no effect since my implementation governs and overrides. I can remove my implementation at any time of my choosing.
3. I am a Block Data API consumer and I use synced patterns. My consuming code explicitly ignores `core/block` blocks since they have no content. Upgrading has no effect.
4. I am a Block Data API consumer and I use synced patterns. My consuming code _effectively_ ignore `core/block` blocks because they currently have no inner blocks. But if they did have inner blocks, my code would render them. Upgrading has an effect, and I am potentially caught off guard by content that is now rendered.

Scenarios 1–3 are not breaking changes. A breaking change via scenario 4 is possible but unlikely IMO.

In summary, we have some theoretically possible breaking changes for you to consider. If we do add a `v2` namespace, we require Block Data API users who want to gain support for these emerging WordPress core features to (1) notice that these features are available (2) update their code to point to the new endpoint.

In addition, we perhaps set a precedent. When this plugin adds support for new core features that alter the _content_—but not the _shape_—of the response, are we obligated to add a new version namespace? Perhaps. With enhancements to block bindings and patterns coming, this question will probably come up again.

alecgeatches:
@chriszarate Good points. I think we should stick with `v1`. In short, given that block bindings are inert with the current Block Data API and unlikely to be relied upon for functionality, this is an improvement that opens new patterns up to future API consumers without disruption.

ingeniumed (approved):
Something minor worth pointing out in our release notes: The order of the attributes has changed in the rest api as well as in the graphQL api. I think it's minor as we shouldn't be worried about that order changing, but just worth noting.

Otherwise, love this! Thanks for working on it @chriszarate

PR #75, opened 2024-08-26
Add synced patterns support

chriszarate:
## Description

Add support for synced patterns (previously known as "reusable blocks"). Synced patterns are implemented as a dynamic block (`core/block`) and currently result in a dead end in the API output:

```json
{
  "blocks": [
    {
      "name": "core/block",
      "attributes": {
        "ref": 6
      }
    }
  ]
}
```

However that `ref` is a pointer to a `wp_block` post, where the blocks (and potential block data) is stored. This PR adds support for loading those blocks:

```json
{
  "blocks": [
    {
      "name": "core/block",
      "attributes": {
        "ref": 6
      },
      "innerBlocks": [
        {
          "name": "core/paragraph",
          "attributes": {
            "content": "This is a block stored in a synced pattern!"
          }
        }
      ]
    }
  ]
}
```

Setting aside tests and comments, the diff here is only about 150 LOC. Its primary features are:

- Add two new actions and one filter that allow the `CoreBlock` class to hook into the parser. These hooks are documented in the README.
- Hook into the core render cycle so that we can capture the inner blocks of synced patterns and inject them as the parser walks the tree.
- Provide support for important core features and use cases, including:
  - Multiple nested synced patterns
  - Block bindings inside synced patterns
  - Synced pattern overrides (itself a special kind of block binding)

The `CoreBlock` does some fairly advanced filtering but is well-documented in the comments. Likewise, the tests are fairly extensive but feel free to suggest additional cases.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1299,6 +1299,38 @@ Note that this filter is evaluated after the [`include`](#include) and [`exclude
 
 ---
 
+### `vip_block_data_api__sourced_block_inner_blocks`
+
+Modify a block's inner blocks before they are recursively added the result tree.
+
+```php
+/**
+ * Filters a block's inner blocks before recursive iteration.
+ *
+ * @param array  $inner_blocks An array of inner block (WP_Block) instances.
+ * @param string $block_name   Name of the parsed block, e.g. 'core/paragraph'.
+ * @param int    $post_id      Post ID associated with the parsed block.
+ * @param array  $block        Result of parse_blocks() for this block.
+ */
+$inner_blocks = apply_filters( 'vip_block_data_api__sourced_block_inner_blocks', $inner_blocks, $block_name, $this->post_id, $block->parsed_block );
+```
+
+This is useful if you want to add or remove inner blocks from the tree based on the parent block. Note that the inner blocks are WP_Block instances, not the associative arrays returned by `parse_blocks`.
+
+```php
+add_filter( 'vip_block_data_api__sourced_block_inner_blocks', 'remove_gallery_inner_blocks', 10, 4 );
+
+function remove_gallery_inner_blocks( $inner_blocks, $block_name, $post_id, $block ) {
+    if ( 'core/gallery' !== $block_name ) {
+        return $inner_blocks;
+    }
+
+    return [];
+}
+```
+
+---
+
 ### `vip_block_data_api__sourced_block_result`
 
 Modify or add attributes to a block's output in the Block Data API.
--- a/README.md
+++ b/README.md
@@ -1321,11 +1321,11 @@ This is useful if you want to add or remove inner blocks from the tree based on 
 add_filter( 'vip_block_data_api__sourced_block_inner_blocks', 'remove_gallery_inner_blocks', 10, 4 );
 
 function remove_gallery_inner_blocks( $inner_blocks, $block_name, $post_id, $block ) {
-    if ( 'core/gallery' !== $block_name ) {
-        return $inner_blocks;
+    if ( 'core/gallery' === $block_name ) {
+        return [];
     }
 
-    return [];
+    return $inner_blocks;
 }
 ```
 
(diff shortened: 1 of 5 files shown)


src/parser/content-parser.php
```diff
@@ -114,13 +115,22 @@ protected function should_block_be_included( WP_Block $block, array $filter_opti
...
+
 		Analytics::record_usage();
 
 		if ( isset( $filter_options['exclude'] ) && isset( $filter_options['include'] ) ) {
 			return new WP_Error( 'vip-block-data-api-invalid-params', 'Cannot provide blocks to exclude and include at the same time', [ 'status' => 400 ] );
 		}
 
-		$this->post_id  = $post_id;
+		// Temporarily set global $post.
+		$previous_global_post = $post;
```
ingeniumed, line 140: Making sure I understand this, why is this needed? Would be worth documenting the reasoning behind this.

ingeniumed (dismissed):
~~Something I noticed that gutenberg does is that, it adds this paragraph block with the text `This is a custom pattern` to every block as an inner block. Should this be something worth filtering out, and offering a filter that could reverse it if someone doesn't want it?~~ Ignore this, as it was me who added a paragraph saying that and forgot about it. Always check your blocks, folks.

It looks good to me. Would want to also get @alecgeatches's take on this PR.

src/parser/content-parser.php
chriszarate: Improved the comment in https://github.com/Automattic/vip-block-data-api/pull/75/commits/0c1b48a52c76d18ed89d998917712dcc36f15ebb, thanks!

ingeniumed (dismissed):

README.md
```diff
@@ -1299,6 +1299,38 @@ Note that this filter is evaluated after the [`include`](#include) and [`exclude
...
+
+```php
+add_filter( 'vip_block_data_api__sourced_block_inner_blocks', 'remove_gallery_inner_blocks', 10, 4 );
+
+function remove_gallery_inner_blocks( $inner_blocks, $block_name, $post_id, $block ) {
+    if ( 'core/gallery' !== $block_name ) {
+        return $inner_blocks;
+    }
+
+    return [];
```
alecgeatches: I know this is just an example and would work, but it feels backwards to me. How about the inverse ordering:

```suggestion
    if ( 'core/gallery' === $block_name ) {
        return [];
    }

    return $inner_blocks;
```

README.md
```diff
@@ -1309,11 +1341,11 @@ Modify or add attributes to a block's output in the Block Data API.
  *
  * @param array  $sourced_block An associative array of parsed block data with keys 'name' and 'attributes'.
  * @param string $block_name    The name of the parsed block, e.g. 'core/paragraph'.
- * @param string $post_id       The post ID associated with the parsed block.
- * @param string $block         The result of parse_blocks() for this block.
+ * @param int    $post_id       The post ID associated with the parsed block.
```
alecgeatches, line 1344: Thank you for fixing these!

src/parser/block-additions/core-block.php
```diff
@@ -0,0 +1,218 @@
...
+				self::$captured_inner_blocks[ $store_key ]['locked'] = true;
+			}
+		}
+
+		// Get the parent block that is currently being rendered. This is fragile,
+		// but is currently the only way we can get access to the parent block from
+		// inside a dynamic block's render callback function.
+		//
+		// https://github.com/WordPress/WordPress/blob/6.6.1/wp-includes/class-wp-block.php#L517
+		$parent_block = WP_Block_Supports::$block_to_render;
```
alecgeatches: Can we wrap this in some sort of safety method in case it becomes private or changes names in the future? Like an `isset()`/`property_exists()` check. I think we can just bail if we don't see the variable we expect, and we'll see that synced pattern tests are failing.

alecgeatches (dismissed):
This works really well! I spent some time reviewing the code and debugging locally to watch how it works.

I left a couple of notes, but because there's nothing major I'll just fix those in this branch and merge. Thank you Chris!

PR #78, opened 2024-09-10
Fix iterator_to_array() error on PHP < 8.2

alecgeatches:
## Description

Fixes https://github.com/Automattic/vip-block-data-api/issues/77, which causes this error:

> ```json
> {
>   "code": "vip-block-data-api-parser-error",
>   "message": "Error parsing post ID 2: iterator_to_array(): Argument #1 ($iterator) must be of type Traversable, array given",
>   "data": {
>     "status": 500
>   }
> }
> ```

This is caused by a call to `iterator_to_array()` in parser code:

https://github.com/Automattic/vip-block-data-api/blob/1593b60b84a91d61845e19c847e8166c1f0e0c6c/src/parser/content-parser.php#L299-L300

[`iterator_to_array()`](https://www.php.net/manual/en/function.iterator-to-array.php) will only accept an array in PHP 8.2.

PHP Changelog for iterator_to_array() showing 8.2 change

Our plugin supports to PHP 8.0, which means 8.0 and 8.1 will currently show this error for any parsed post.

The changed code now tests for an array and uses it directly if available, which will work on prior PHP versions.

## Steps to Test

1. Check out the `trunk` branch of `vip-block-data-api` and install on a WordPress installation with PHP version `8.0` or `8.1`.
2. Go to a published post's Block Data API URL, e.g.

    ```
    https://my.lndo.site/wp-json/vip-block-data-api/v1/posts/1/blocks
    ```

3. See the iterator error in the response data: `Error parsing post ID 1: iterator_to_array(): Argument #1 ($iterator) must be of type Traversable, array given`.
4. Change the VIP Block Data API branch to this branch (`fix/php-8.1-iterator-error`).
5. Reload the `wp-json` endpoint used in step 2.
6. See that the post renders correctly.

## Other Todos

- [x] ~Figure out why the [minimum test matrix](https://github.com/Automattic/vip-workflow-plugin/blob/b45b565b49be4c21d5511cb79e927c06c463bc83/.github/workflows/php-tests.yml#L20-L23) (PHP 8/WordPress 6.2) didn't catch this.~

    See https://github.com/Automattic/vip-block-data-api/pull/79.
- [x] Ensure this fix is included in the next update of VIP's integrations.

PR #79, opened 2024-09-10
Fix wp-env to use proper PHP version in test matrix

alecgeatches:
## Description

See https://github.com/Automattic/vip-block-data-api/pull/78, which addresses a bug in `1.4.0` that should have caused test failures due to a regression. The cause of the issue is that our test matrix PHP version was being used only partially to setup depedencies and run composer:

https://github.com/Automattic/vip-block-data-api/blob/1593b60b84a91d61845e19c847e8166c1f0e0c6c/.github/workflows/phpunit.yml#L36-L40

but not explicitly passed to `wp-env`:

https://github.com/Automattic/vip-block-data-api/blob/1593b60b84a91d61845e19c847e8166c1f0e0c6c/.github/workflows/phpunit.yml#L47-L50

This PR addresses these issues:

- We now pass the WordPress PHP version via `WP_ENV_PHP_VERSION`.
- We've added an additional test test with PHP 8.0 and the latest version of WordPress, which may catch similar PHP-related changes in recent versions of WordPress features.
- The PHP "latest" option [doesn't work](https://github.com/Automattic/vip-block-data-api/actions/runs/10798196163/job/29951165023#step:6:13) with `wp-env` as an option for `WP_ENV_PHP_VERSION`:

    ```
    ✖ Invalid environment variable: "WP_ENV_PHP_VERSION" must be a string of the format "X", "X.X", or "X.X.X".
    ```

    We're passing `8.3` explicitly instead. We'll want to keep this up to date with newer PHP versions, but this is a helpful guide for our highest tested version, which we did not explicitly test before.

### DIFF
--- a/.github/workflows/phpunit.yml
+++ b/.github/workflows/phpunit.yml
@@ -21,6 +21,10 @@ jobs:
           - php: "8.0"
             wp: "6.0"
             allowed_failure: false
+          # Check latest WP with the lowest supported PHP.
+          - php: "8.0"
+            wp: "master"
+            allowed_failure: false
           # Check latest WP with the highest supported PHP.
           - php: "latest"
             wp: "master"
--- a/.github/workflows/phpunit.yml
+++ b/.github/workflows/phpunit.yml
@@ -52,6 +52,7 @@ jobs:
         run: wp-env start
         env:
           WP_ENV_CORE: WordPress/WordPress#${{ matrix.wp }}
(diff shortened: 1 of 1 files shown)


alecgeatches:
I'm going to merge this directly to test that the changes in https://github.com/Automattic/vip-block-data-api/pull/78 fix the currently-broken tests. This is also a test-only change, so risk is very low.

PR #80, opened 2024-09-12
Add min php and wp notice

ingeniumed:
## Description

I realized we don't have a PHP and WP version check and the one we have been relying won't really be enforced for the way we expect this plugin to be deployed. So I have added in a manual check, akin to what we have in the workflow plugin.

This is already what the plugin supports so it's just matching that.

## Steps to Test

Either load a WP version prior to 6.0 with PHP version lower than 8.0, or change those numbers to something arbitrary and see it fail.

### DIFF
--- a/vip-block-data-api.php
+++ b/vip-block-data-api.php
@@ -5,7 +5,7 @@
  * Description: Access Gutenberg block data in JSON via the REST API.
  * Author: WordPress VIP
  * Text Domain: vip-block-data-api
- * Version: 1.4.1
+ * Version: 1.4.2
  * Requires at least: 6.0
  * Tested up to: 6.6
  * Requires PHP: 8.0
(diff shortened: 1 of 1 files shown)


sanmai (approved):

alecgeatches (approved):
Changes look good to me, but I'm not sure that we need to bump the version and do a release for this PR alone. What about leaving it at `1.4.1`, and then this will get merged when we do a next release?

ingeniumed:
> Changes look good to me, but I'm not sure that we need to bump the version and do a release for this PR alone. What about leaving it at `1.4.1`, and then this will get merged when we do a next release?

Agreed on that, works for me tbh

alecgeatches (approved):
Merged #82 and #83, going to move this in as well. Thank you!

Issue #81, opened 2024-10-17
Incorrect Handling of Multiple DOMElement Objects in source_block_raw Method


veerap000:
I found an issue with the `source_block_raw` method in [/src/parser/content-parser.php](https://github.com/Automattic/vip-block-data-api/blob/trunk/src/parser/content-parser.php#L677).

### Current Behavior:

In the current implementation, the function only returns the first matched DOMElement when there are multiple elements within the _core/html_ block. Specifically, it uses:
`$attribute_value = trim( $crawler->outerHtml() );`

This causes a problem when the _core/html_ block contains multiple DOMElement objects. For example, consider the following HTML content:
`<h2>My HTML heading</h2><p>My HTML paragraph</p>`

Currently, only the first element (`<h2>My HTML heading</h2>`) is returned, while the rest of the elements are ignored.

### Proposed Solution:

To address this, I propose modifying the function to iterate over all matched elements, collect their outer HTML, and then concatenate them together. This can be accomplished with the following change:

```
$attribute_values = $crawler->each(function ($node) {
    return trim( $node->outerHtml() );
});

$attribute_value = trim( implode( '', $attribute_values ) );
```

This approach ensures that all DOMElement objects within the core/html block are captured and returned correctly.

**Example Output:**
With this change, the example input:
`<h2>My HTML heading</h2><p>My HTML paragraph</p>`
Will be correctly handled, returning both elements as expected, instead of only the first.

Please let me know if this proposed fix is suitable or if further adjustments are needed.

alecgeatches:
@veerap000 Thank you for the report! I think your problem explanation and solution look right to me. We'll see if we can get this fixed shortly.

alecgeatches:
@veerap000 Merged a fix over in https://github.com/Automattic/vip-block-data-api/pull/82! We used a slightly different approach:

https://github.com/Automattic/vip-block-data-api/blob/a4939bbe953548544e670e74207d1973d1ab4ae0/src/parser/content-parser.php#L685-L694

I noticed during testing that the `trim()` + `implode()` method would remove inner whitespace, which can be important for text like this:

```html
<span>These</span> <span class="highlight">spaces</span> <span>are</span> <span>important</span>.
```

[...]

PR #82, opened 2024-10-18
Fix raw attribute sources with multiple nodes

alecgeatches:
## Description

See https://github.com/Automattic/vip-block-data-api/issues/81 for a description of the problem. In short, if a block with a raw source like `core/html` has multiple top-level nodes, the Block Data API returns only the HTML of the first node.

Block HTML:

```
<!-- wp:html -->
<h1>First node</h1><h2>Second node</h2>
<!-- /wp:html -->
```

Incorrect output:

```json
{
    "name": "core/html",
    "attributes": {
        "content": "<h1>First node</h1>"
    }
}
```

The `source_block_raw` function responsible for gathering this HTML has been changed to use the parent wrapper block's inner HTML. This will collect all node HTML as well as preserving inner whitespace in the HTML content.

## Steps to Test

1. Check out PR.
1. Run `wp-env start` and `composer install`.
1. Run `composer run test`
1. See successful tests.

### DIFF
--- a/src/parser/content-parser.php
+++ b/src/parser/content-parser.php
@@ -680,10 +680,15 @@ class ContentParser {
 		// Also see tag attribute parsing in Gutenberg:
 		// https://github.com/WordPress/gutenberg/blob/6517008/packages/blocks/src/api/parser/get-block-attributes.js#L131
 
-		$attribute_value = null;
-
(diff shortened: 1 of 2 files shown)


alecgeatches:
Merging! Test failures are from block bindings differences in latest WordPress, unrelated to the changes here.

PR #87, opened 2025-03-20
Correctly parse synced patterns with nested blocks

### LINKED ISSUE #86: Nested blocks in a synced pattern are listed multiple times
**Describe the bug**
When a nested block structure is put in a synced pattern (reusable block), some of the blocks in the resulting array are listed multiple times.

**To Reproduce**
- Create a synced pattern with the following content: 
```
			<!-- wp:group -->
			<div class="wp-block-group">
				<!-- wp:group -->
				<div class="wp-block-group">
					<!-- wp:paragraph -->
					<p>p1</p>
					<!-- /wp:paragraph -->
				</div>
				<!-- /wp:group -->
			</div>
			<!-- /wp:group -->
```
- Reference the above pattern into a `<!-- wp:block {"ref":%d} /-->` in a separate post and check the Block API output.

**Expected behavior**
I would expect to receive the following (this is PHP from the test I wrote)
```
[
	[
		'name'        => 'core/block',
		'attributes'  => [
			'ref' => $synced_pattern->ID,
		],
		'innerBlocks' => [
			[
				'name'        => 'core/group',
				'attributes'  => [
					'tagName' => 'div',
				],
				'innerBlocks' => [
					[
						'name'        => 'core/group',
						'attributes'  => [
							'tagName' => 'div',
						],
						'innerBlocks' => [
							[
								'name'       => 'core/paragraph',
								'attributes' => [
									'content' => 'p1',
									'dropCap' => false,
								],
							],
						],
					],
				],
			],
		],
	],
]
```

**Actual behavior**
Instead, I'm getting the following:
```
[
  {
    "name": "core/block",
    "attributes": {
      "ref": 4
    },
    "innerBlocks": [
      {
        "name": "core/paragraph",
        "attributes": {
          "content": "p1",
          "dropCap": false
        }
      },
      {
        "name": "core/group",
        "attributes": {
          "tagName": "div"
        },
        "innerBlocks": [
          {
            "name": "core/paragraph",
            "attributes": {
              "content": "p1",
              "dropCap": false
            }
          }
        ]
      },
      {
        "name": "core/group",
        "attributes": {
          "tagName": "div"
        },
        "innerBlocks": [
          {
            "name": "core/group",
            "attributes": {
              "tagName": "div"
            },
            "innerBlocks": [
              {
                "name": "core/paragraph",
                "attributes": {
                  "content": "p1",
                  "dropCap": false
                }
              }
            ]
          }
        ]
      }
    ]
  }
]
```
So the `core/paragraph` block is listed multiple times, in all of its parent blocks. Same for the second `core/group`. This is different from the output you'd get from the Block API if the blocks in the reusable block structure were used directly and not through a `core/block`. 

**Version of the plugin**
1.4.3

chriszarate:
## Description

Fixes https://github.com/Automattic/vip-block-data-api/issues/86.

Our current solution for extracting and parsing blocks from synced patterns is a bit too experimental. It hooks into the render cycle and attempts to detect when the block being rendered has a synced pattern parent block. It does this via a somewhat fragile inspection of `WP_Block_Supports`.

The issue with that approach is that it can't tell when the synced pattern is a direct parent block vs. an ancestor block, so it ends up duplicating blocks in nested trees. There's no way to make this determination reliably, so we need to abandon that approach. 

Instead of hooking into the existing render cycle, we can just render synced patterns manually. This carries some of its own fragility, because we need to duplicate some code from Core to support synced pattern overrides. But it's an overall improvement and the duplicated code is just a few lines.

It does require making `ContentParser#render_parsed_block` a public method, which was the path I found that required the least refactoring.

@toshotosho Thanks for the failing test case. Can you check out this branch and make sure it resolves the issue for you? Feel free to contribute additional test cases.

### DIFF
--- a/src/parser/block-additions/core-block.php
+++ b/src/parser/block-additions/core-block.php
@@ -10,10 +10,11 @@ namespace WPCOMVIP\BlockDataApi\ContentParser\BlockAdditions;
 defined( 'ABSPATH' ) || die();
 
 use WP_Block;
-use WP_Block_Supports;
-use function add_action;
+use WPCOMVIP\BlockDataApi\ContentParser;
+
 use function add_filter;
-use function remove_filter;
+use function get_post;
+use function parse_blocks;
 
 /**
  * Enhance core/block block by capturing its inner blocks.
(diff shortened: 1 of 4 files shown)


alecgeatches (dismissed):
Great fix! I like that the `core/block` addition is actually smaller now, and we've doubled the size of the tests on synced patterns. The new tests cover a lot more.

It'd be great to have @toshotosho verify on real data first per the bug in https://github.com/Automattic/vip-block-data-api/issues/86, but I'm happy to merge when verified.

tests/parser/test-synced-patterns.php
```diff
@@ -226,6 +226,273 @@ public function test_synced_pattern_with_override() {
...
+		$this->assertArrayHasKey( 'blocks', $blocks, sprintf( 'Unexpected parser output: %s', wp_json_encode( $blocks ) ) );
+		$this->assertEquals( $expected_blocks, $blocks['blocks'], sprintf( 'Blocks not equal: %s', wp_json_encode( $blocks['blocks'] ) ) );
+	}
+
+	/* Synced pattern with override in nested content */
+
+	public function test_synced_pattern_with_override_in_nested_content() {
+		$synced_pattern_content = '
+			<!-- wp:group -->
+			<div class=k"wp-block-group">
```
alecgeatches: Minor note, there's 3 `k"`s scattered in here that appear to be an accident, unless I'm missing something.

chriszarate: Fixed in https://github.com/Automattic/vip-block-data-api/pull/87/commits/5dbe30812dd750bfd8414e184b97c06dff687776

#vimfail

src/parser/block-additions/core-block.php
```diff
@@ -88,105 +52,47 @@ public static function cleanup_after_render() {
...
-	 * @param WP_Block $block          Inner block.
-	 */
-	protected static function capture_inner_block( array $synced_pattern, WP_Block $block ): void {
-		$store_key = self::get_store_key( $synced_pattern );
-		if ( ! isset( self::$captured_inner_blocks[ $store_key ] ) ) {
-			self::$captured_inner_blocks[ $store_key ] = [
-				'inner_blocks' => [],
-				'locked'       => false,
-			];
+		if ( ! $post instanceof \WP_Post ) {
```
alecgeatches: Extreme nit - can we use `use WP_Block;` up top and this for consistency?

```suggestion
		if ( ! $post instanceof WP_Post ) {
```

chriszarate: Nice catch! Fixed in https://github.com/Automattic/vip-block-data-api/pull/87/commits/fc2457531a99621d2d8ca62c1a2728df8f0e7fbf

toshotosho:
Thanks guys!

I just deployed the fix and it solves our issue without breaking anything else, so great work!

alecgeatches (approved):

PR #90, opened 2025-11-19
Updating to IC installation.

smithjw1:
## Description

Updates to the readme with better install instructions on VIP.

## Steps to Test

N/A


ingeniumed (approved):
Should update the documentation in the wpvip docs as well

PR #93, opened 2025-12-15
Parser: Allow empty content

### LINKED ISSUE #92: Block Data API throws error when there is no block content
**Describe the bug**
When querying a post in GraphQL with no block content, even when using the Gutenberg editor, [this block-data-api code](https://github.com/Automattic/vip-decoupled-bundle/blob/trunk/lib/vip-block-data-api/src/parser/content-parser.php#L151-L158) causes an error to be thrown. It's ostensibly to check for classic editor content but will also throw an error when there is no block content (causing `$has_blocks` to return false and trigger the error). I would expect it to return an empty array for block content, not error out.

**To Reproduce**
- Create a post with no blocks and query the post with:
```
query NewQuery {
  post(id: 1, idType: DATABASE_ID) {
    blocksDataV2 {
      blocks {
        name
        id
        parentId
        attributes {
          name
          value
        }
      }
    }
  }
}
```

**Expected behavior**
A data object for the post with `blocks` as an empty array should be returned:
```
"post": {
      "blocksDataV2": {
        "blocks": []
      }
    }
```

**Actual behavior**
Internal server error is returned, e.g.: `vip-block-data-api-no-blocks - Error parsing post ID 1273: This post does not appear to contain block content. The VIP Block Data API is designed to parse Gutenberg blocks and can not read classic editor content. in /wp/wp-content/mu-plugins/vip-integrations/vip-block-data-api-1.4/src/analytics/analytics.php on line 58`

**Version of the plugin**
1.4.5 (activated with `\Automattic\VIP\Integrations\activate( 'block-data-api' );` in `plugin-loader.php`)

**Additional context**
We have a use case where custom post type pages are largely generated from external data sources, but it's possible to create pages for them in WordPress to allow for overrides of default options and/or add additional content with Gutenberg blocks. Many times users will only utilize custom meta fields with no need for actual blocks of content to be added to the Gutenberg editor.

chriszarate:
## Description

Empty content fails the `has_blocks` check, but is valid block markup. Fixes #92 

## Steps to Test

Covered by new unit tests.

### DIFF
--- a/src/parser/content-parser.php
+++ b/src/parser/content-parser.php
@@ -149,7 +149,9 @@ class ContentParser {
 		// An empty `post_content` fails `has_blocks` check, but is still valid.
 		$content_is_empty = empty( trim( $post_content ) );
 
-		if ( ! ( $content_is_empty || has_blocks( $post_content ) ) ) {
+		$has_content = ! empty( trim( $post_content ) );
+		
+		if ( $has_content && ! has_blocks( $post_content ) ) {
 			$error_message = join(' ', [
 				sprintf( 'Error parsing post ID %d: This post does not appear to contain block content.', $post_id ),
 				'The VIP Block Data API is designed to parse Gutenberg blocks and can not read classic editor content.',
--- a/src/parser/content-parser.php
+++ b/src/parser/content-parser.php
@@ -146,9 +146,10 @@ class ContentParser {
 
 		$this->warnings = [];
 
-		$has_blocks = has_blocks( $post_content );
+		// An empty `post_content` fails `has_blocks` check, but is still valid.
+		$content_is_empty = empty( trim( $post_content ) );
 
-		if ( ! $has_blocks ) {
+		if ( ! ( $content_is_empty || has_blocks( $post_content ) ) ) {
 			$error_message = join(' ', [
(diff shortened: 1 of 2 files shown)


src/parser/content-parser.php
```diff
@@ -146,9 +146,10 @@ public function parse( $post_content, $post_id = null, $filter_options = [] ) {
 
 		$this->warnings = [];
 
-		$has_blocks = has_blocks( $post_content );
+		// An empty `post_content` fails `has_blocks` check, but is still valid.
+		$content_is_empty = empty( trim( $post_content ) );
 
-		if ( ! $has_blocks ) {
+		if ( ! ( $content_is_empty || has_blocks( $post_content ) ) ) {
```
alecgeatches: Minor suggestion, the original conditional here took me a few attempts to understand

```suggestion
		$has_content = ! empty( trim( $post_content ) );
		
		if ( $has_content && ! has_blocks( $post_content ) ) {
```

alecgeatches (dismissed):
Thank you for making this fix and adding more tests to cover it. Minor suggestion above, feel free to merge when you've taken a look and I can put out a new release.

chriszarate:
@alecgeatches Updated. If good, feel free to merge and release

alecgeatches (approved):
Thank you @chriszarate!

Issue #47, opened 2023-08-30
Automate the release process


ingeniumed:
We have too many manual steps for creating a release in RELEAE.md, and we should automate that. Gutenberg can be used as an example for how to do this.

alecgeatches:
Completed in https://github.com/Automattic/vip-block-data-api/pull/94! We just had our first automated release with https://github.com/Automattic/vip-block-data-api/releases/tag/v1.4.6.

PR #94, opened 2025-12-31
Add release workflow automation

alecgeatches:
## Description

See https://github.com/Automattic/vip-block-data-api/issues/47. Recreate the steps in [RELEASE.md](https://github.com/Automattic/vip-block-data-api/blob/ddc91c93d7d316fe9255693da60da38761651a96/RELEASE.md) in a GitHub workflow. After this is merged, we should be able to update the plugin version on `trunk` and the tag/zip should be auto-created with release notes.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -4,6 +4,9 @@ on:
     branches:
       - trunk
 
+permissions:
+  contents: write
+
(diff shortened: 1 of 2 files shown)


alecgeatches:
@copilot Does this look like it'll work?

PR #100, opened 2026-05-08
fix: allow null block names in core block filter

### LINKED ISSUE #99: TypeError: CoreBlock::get_inner_blocks() argument #2 ($block_name) must be of type string, null given — when parsing posts that contain blocks with a null blockName
### Describe the bug
 
`WPCOMVIP\BlockDataApi\ContentParser\BlockAdditions\CoreBlock::get_inner_blocks()` declares its second parameter as a non-nullable `string $block_name`. Under PHP 8 strict types, this method throws a `TypeError` when invoked via the `vip_block_data_api__sourced_block_inner_blocks` filter for blocks whose parsed `blockName` is `null` — which `parse_blocks()` legitimately emits for non-block content (raw HTML, pasted classic-editor content, or any text fragment sitting between block delimiters).
 
The `TypeError` is caught by `ContentParser::parse()`'s `Throwable` handler and returned as a `WP_Error` with `status => 500`, which the REST API surfaces as HTTP 500 to the caller. From the consumer's perspective, parsing the affected post returns 500 every time. Because this parsing path is used by any consumer of the Block Data API, both editor REST traffic and any server-side rendering pipeline that relies on the parser are affected.
 
The same non-nullable signature exists on `CoreBlock::remove_content_array()`, which is registered against the `vip_block_data_api__sourced_block_result` filter and would fail in identical conditions.
 
> Note: whitespace-only blocks (null `blockName` + empty trimmed `inner_html`) are correctly excluded earlier in `ContentParser::should_block_be_included()`. The trigger is specifically blocks with `blockName === null` *and* non-empty `inner_html`.
 
### Exception
 
```
WPCOMVIP\BlockDataApi\ContentParser\BlockAdditions\CoreBlock::get_inner_blocks():
Argument #2 ($block_name) must be of type string, null given,
called in /var/www/wp-includes/class-wp-hook.php on line 324
```
 
The same exception fires consistently for affected post IDs, recurring whenever those posts are read through Block Data API.
 
### Affected method
 
[`src/parser/block-additions/core-block.php`](https://github.com/Automattic/vip-block-data-api/blob/trunk/src/parser/block-additions/core-block.php)
 
```php
public static function get_inner_blocks(
    array $inner_blocks,
    string $block_name,            // <-- this should be ?string
    int|null $post_id,
    array $parsed_block
): array {
    // Not a synced pattern? Return the inner blocks unchanged.
    if ( self::$block_name !== $block_name || ! isset( $parsed_block['attrs']['ref'] ) ) {
        return $inner_blocks;
    }
    ...
}
```
 
### Reproduction
 
**Programmatic:**
 
```php
$content = "<!-- wp:paragraph --><p>hello</p><!-- /wp:paragraph -->\n\n<p>orphan paragraph</p>";
$parser  = new \WPCOMVIP\BlockDataApi\ContentParser();
$result  = $parser->parse( $content, 0 );
// $result is a WP_Error with status 500 and message:
// "Error parsing post ID 0: WPCOMVIP\BlockDataApi\ContentParser\BlockAdditions\CoreBlock::get_inner_blocks():
//  Argument #2 ($block_name) must be of type string, null given..."
```
 
**Via REST:**
 
1. Create a post whose content includes at least one block delimiter (so it passes the `has_blocks()` check) and at least one non-block fragment, e.g.:
   ```html
   <!-- wp:paragraph --><p>hello</p><!-- /wp:paragraph -->
 
   <p>orphan paragraph not wrapped in a block</p>
   ```
 
2. Request the post via any endpoint that invokes `ContentParser::parse()`.
3. Observe HTTP 500 with the `WP_Error` message above.
The orphan paragraph produces an entry in `parse_blocks()` output with `blockName: null` and non-empty `innerHTML`, which is what triggers the filter call with `$block_name = null`.
 
### Suggested fix
 
The early-return guard already handles `null` correctly (`'core/block' !== null` is true), so the type signature can simply be made nullable:
 
```diff
 public static function get_inner_blocks(
     array $inner_blocks,
-    string $block_name,
+    ?string $block_name,
     int|null $post_id,
     array $parsed_block
 ): array {
```
 
The same nullable change is needed on `CoreBlock::remove_content_array()` for the `vip_block_data_api__sourced_block_result` filter — same root cause, same fix.
 
### Impact
 
In a high-traffic, editorially active WordPress site, this manifests as recurring spikes of HTTP 500s clustered around editorial REST endpoints (e.g. autosave, wp/v2 sourcing endpoints, custom REST routes that consume Block Data API) and any SSR pipeline that uses the parser. The errors are non-deterministic from the user's perspective — they only fire for posts that happen to contain a null-blockName block with non-empty content, which is easy to introduce accidentally via classic-to-block migrations or pasted HTML.
 
### Environment
 
- `vip-block-data-api` version: 1.4.7 (current trunk; the file at `src/parser/block-additions/core-block.php` on `trunk` still has the non-nullable signature at the time of filing)
- PHP version: 8.3
- WordPress version: 6.8.5

alceops:
Fixes #99.

This updates the synced-pattern inner-block filter to accept the null block names that WordPress can pass for parsed non-block/classic-content fragments. In that case the filter now returns the existing inner blocks unchanged instead of raising a PHP TypeError before the parser can continue.

I also added a focused regression test that calls the filter with a null block name.

Verification: I could not run the WordPress/PHP test suite in this cron profile because php/composer/git are not installed here. I did a static patch check against the current trunk file contents via the GitHub API.

### DIFF
--- a/src/parser/block-additions/core-block.php
+++ b/src/parser/block-additions/core-block.php
@@ -58,12 +58,12 @@ class CoreBlock {
 	 * https://github.com/WordPress/WordPress/blob/6.6.1/wp-includes/blocks/block.php#L19
 	 *
 	 * @param array    $inner_blocks Inner blocks.
-	 * @param string   $block_name   Block name.
+	 * @param string|null $block_name   Block name, or null for parsed non-block content.
 	 * @param int|null $post_id      Post ID.
 	 * @param array    $parsed_block Parsed block data.
 	 * @return array
 	 */
-	public static function get_inner_blocks( array $inner_blocks, string $block_name, int|null $post_id, array $parsed_block ): array {
+	public static function get_inner_blocks( array $inner_blocks, string|null $block_name, int|null $post_id, array $parsed_block ): array {
 		// Not a synced pattern? Return the inner blocks unchanged.
 		if ( self::$block_name !== $block_name || ! isset( $parsed_block['attrs']['ref'] ) ) {
 			return $inner_blocks;
--- a/tests/parser/test-inner-blocks.php
+++ b/tests/parser/test-inner-blocks.php
@@ -186,4 +186,17 @@ class InnerBlocksTest extends RegistryTestCase {
 		$this->assertArrayHasKey( 'blocks', $blocks, sprintf( 'Unexpected parser output: %s', wp_json_encode( $blocks ) ) );
 		$this->assertArraySubset( $expected_blocks, $blocks['blocks'], true );
 	}
+	public function test_core_block_inner_blocks_filter_ignores_null_block_names() {
+		$inner_blocks = [ 'existing-inner-block' ];
+
+		$result = \WPCOMVIP\BlockDataApi\ContentParser\BlockAdditions\CoreBlock::get_inner_blocks(
+			$inner_blocks,
+			null,
+			1,
+			[ 'blockName' => null ]
+		);
+
+		$this->assertSame( $inner_blocks, $result );
+	}
+
 }


alecgeatches:
@alceops Thank you for the fix! I'm happy to merge, but can you address the `phpcs` failures [in this run](https://github.com/Automattic/vip-block-data-api/actions/runs/25551385398/job/75423534875?pr=100) please? Note that the workflow didn't run automatically since you're a first-time contributor. Thank you!

```
FILE: tests/parser/test-inner-blocks.php
--------------------------------------------------------------------------------
FOUND 1 ERROR AFFECTING 1 LINE
--------------------------------------------------------------------------------
 202 | ERROR | [x] The closing brace for the class must go on the next line
     |       |     after the body
     |       |     (PSR2.Classes.ClassDeclaration.CloseBraceAfterBody)
--------------------------------------------------------------------------------
PHPCBF CAN FIX THE 1 MARKED SNIFF VIOLATIONS AUTOMATICALLY
--------------------------------------------------------------------------------


FILE: src/parser/block-additions/core-block.php
--------------------------------------------------------------------------------
FOUND 3 ERRORS AFFECTING 3 LINES
--------------------------------------------------------------------------------
 60 | ERROR | [x] Expected 7 spaces after parameter type; 4 found
    |       |     (Squiz.Commenting.FunctionComment.SpacingAfterParamType)
 62 | ERROR | [x] Expected 4 spaces after parameter type; 1 found
    |       |     (Squiz.Commenting.FunctionComment.SpacingAfterParamType)
 63 | ERROR | [x] Expected 7 spaces after parameter type; 4 found
    |       |     (Squiz.Commenting.FunctionComment.SpacingAfterParamType)
--------------------------------------------------------------------------------
```

alecgeatches:
I opened up a separate PR to address the `phpcs` issues and merged here: https://github.com/Automattic/vip-block-data-api/pull/102. Thank you for this pull request, will make a new release shortly!

alecgeatches:
This (and the text fixes in #102) have been merged into the latest `1.4.8` release: https://github.com/Automattic/vip-block-data-api/releases/tag/1.4.8.

Issue #115, opened 2026-09-18
core/video always returns controls true, even when controls are turned off.


codedbyglenden:
**Describe the bug**
Turning the controls on a core/video off still returns controls:true in the json endpoint. 

**To Reproduce**
- Create a core/video
- Set controls toggle to off
- View the endpoint provided by the plugin, controls still show true.

**Expected behavior**
When toggled off controls should return false.

**Actual behavior**
The json falsely reports that the video controls for this block should be turned on.

**Block Data API URL**

If publicly accessible, please share the block data API URL used to reproduce this problem.

e.g. `https://my.site/wp-json/vip-block-data-api/v1/posts/139/blocks`

**Version of the plugin**
1.4.10

ingeniumed:
Thanks for reporting this! Closing this as completed by https://github.com/Automattic/vip-block-data-api/pull/116.

Issue #114, opened 2026-09-17
CoreBlock::remove_content_array() rejects the null block name the parser passes, failing the whole parse


Zmetser:
## Summary

`CoreBlock::remove_content_array()` declares `string $block_name`, but the filter it is registered on passes `null` for content that sits outside block delimiters. The resulting TypeError is caught as a `WP_Error`, and the whole parse fails.

A consumer loses **every** block on the post, not one.

Observed on vip-block-data-api 1.4.10:

```
WPCOMVIP\BlockDataApi\ContentParser\BlockAdditions\CoreBlock::remove_content_array():
Argument #2 ($block_name) must be of type string, null given
```

## The plugin already documents that the name is nullable

Both methods sit in `src/parser/block-additions/core-block.php` and are registered on adjacent lines:

```php
37: add_filter( 'vip_block_data_api__sourced_block_inner_blocks', [ __CLASS__, 'get_inner_blocks' ], 5, 5 );
38: add_filter( 'vip_block_data_api__sourced_block_result',      [ __CLASS__, 'remove_content_array' ], 5, 2 );
```

One accepts null. The other does not:

```php
61:  * @param string|null $block_name   Block name, or null for parsed non-block content.
66:  public static function get_inner_blocks( array $inner_blocks, string|null $block_name, int|null $post_id, array $parsed_block ): array

107: * @param string $block_name    Block name.
110: public static function remove_content_array( array $sourced_block, string $block_name ): array
```

The docblock on line 61 states the case explicitly: "or null for parsed non-block content".

## Why a null name reaches the filter

`ContentParser::should_block_be_included()` drops a null-name block only when it is whitespace-only:

```php
// Whitespace blocks are always excluded.
$is_whitespace_block = null === $block_name && empty( trim( $block->inner_html ) );
if ( $is_whitespace_block ) {
    return false;
}
```

A null-name block that holds real content is therefore included. It reaches `source_block()`, which fires `vip_block_data_api__sourced_block_result`, which calls `remove_content_array()` with a null name.

## Reproduce

1. Create a post whose content mixes blocks with classic text that sits outside any block delimiter. Content imported from the classic editor produces this shape.
2. Parse the post.

The parse returns a `WP_Error` carrying the TypeError above. A REST consumer receives no parsed block data for the post at all.

## Impact

The failure is silent and total for the affected post. Nothing warns, and the post's markup stays intact everywhere else, so a headless consumer simply renders an empty page body.

[...]

PR #116, opened 2026-09-22
Fix block parsing regressions and update minimum requirements

### LINKED ISSUE #113: core/html content is dropped since WordPress removed "source": "raw" from the block
## Summary

Every `core/html` block reaches API consumers with no content. WordPress removed `"source": "raw"` from the block's `content` attribute, and the parser has no path that handles the new shape.

Observed with vip-block-data-api 1.4.10 on WordPress 7.1.

## What changed upstream

Gutenberg commit [`50ddbc1f6`](https://github.com/WordPress/gutenberg/commit/50ddbc1f6), PR #79115, "Blocks: Add innerContent support for static inner blocks, adopt it in the HTML block", 2026-06-29. The patch on `packages/block-library/src/html/block.json`:

```diff
 "content": {
   "type": "string",
-  "source": "raw",
-  "role": "content"
+  "role": "local"
 }
```

The attribute now declares no `source`.

## Why the parser drops the value

In `ContentParser::apply_sourced_attributes()` the source is read with `$attribute_source = $definition['source'] ?? null`. That is now null, so the code takes the branch that assumes the value sits in the block delimiter attributes.

`core/html` keeps its content in the block's inner HTML, not in the delimiter. The attribute also declares no `default`. So the branch reaches its final `continue`, the DOM crawler is never built, and `source_attribute()` is never called.

The block is then emitted with empty attributes:

```json
{ "name": "core/html", "attributes": {} }
```

## Reproduce

1. Run WordPress 7.1 with vip-block-data-api 1.4.10.
2. Add a Custom HTML block to a post, with any markup in it.
3. Request the post through the REST API and read the block in the parsed output.

The block appears with an empty `attributes` object. The markup is intact in `post_content` and in `content.rendered`. Only the parsed output loses it.

Sourcing is otherwise healthy. On the same post in the same parse, `core/paragraph` returns its `content` correctly, because `rich-text` is still declared.

## Impact

A consumer that renders from the parsed block data shows nothing for the block. Nothing errors and nothing warns, so the failure is silent. A headless site loses authored content from published pages.

## Suggested fix

Handle the new mechanism. Either:

1. Read the block's inner HTML for an attribute that declares `role: local` and no `source` on `core/html`, or
2. Support the `innerContent` mechanism that PR #79115 introduced.

## Also, a stale comment

`source_block_raw()` states:

```php
// The only current usage of the 'raw' attribute in Gutenberg core is the 'core/html' block:
```

That is no longer true. `core/html` no longer uses it. The function stays live for `core/shortcode`, `core/freeform` and `core/missing`, which still declare `source: raw`, so the comment should name those instead.

ingeniumed:
## Description

Fixes [#113](https://github.com/Automattic/vip-block-data-api/issues/113), [#114](https://github.com/Automattic/vip-block-data-api/issues/114), and [#115](https://github.com/Automattic/vip-block-data-api/issues/115).

| Before | After |
| --- | --- |
| Custom HTML blocks using the newer `role: local` definition lose their content. | `core/html` returns its inner HTML as `content`; older `raw` definitions remain supported. |
| Classic content between blocks can cause a `TypeError` and fail the entire parse. | Mixed content parses successfully without a misleading unregistered-block warning. |
| A video with controls turned off returns `controls: true`. | Boolean HTML attributes reflect presence, so an absent `controls` attribute returns `false`. |

Raise the minimum requirements to WordPress 6.8 and PHP 8.2. Align the WordPress test dependency and extend CI coverage through PHP 8.5. The existing Symfony 6.4 runtime versions remain in place.

## Steps to Test

1. `WP_ENV_CORE=WordPress/WordPress#6.8 WP_ENV_PHP_VERSION=8.2 wp-env start`
2. `composer test`
3. `composer test-multisite`
4. `composer phpcs`


alecgeatches (approved):
Tested and these changes work for me. Thanks!

